PR #25, opened 2015-06-25
Re: #232 - unit test changes

andyjmaclean:


rwd:
@andyjmaclean:
* The changes to spec_helper.rb in are unnecessary because webmock_helper.rb in develop already has them
* We don't want the local non-headless testing stuff in capybara_helper.rb merged because it won't run on Travis

PR #27, opened 2015-06-28
Feat/267

andyjmaclean:
This is the latest done on #267, already merged with Dan's work on #249, including the minor changes needed to finish work on #260.


rwd:
@andyjmaclean please clean up the trailing whitespace style issues reported by Hound, above. Ignore the double-quote reports as we are favouring single quotes but I haven't figured out how to tell Hound that yet.

PR #29, opened 2015-07-01
Refactor directory structure to store templates in app/views

rwd:
At present we have Mustache templates in the styleguide gem and any overrides in the application stored in the directory app/ui-components, and view classes in app/views. We have two directories as this is what the stache gem defaults to.

I propose that we move the templates into the app/views directory. Any overriden templates in the application would then be in the same directory as their view class, e.g. app/views/templates/search/search_home.rb and app/views/templates/Search/Search-home.mustache

This is for a number of reasons:

1. First and foremost, Mustache view specs do not work if the view classes and templates are in different directories
2. "ui-components" is ambiguous, i.e. all images, stylesheets and javascript are also UI components
3. "app/views" is the Rails convention
4. Storing view classes and templates in the same directory is the mustache gem's convention


rwd:
The only broken test here is the one discussed in #28, so this can be merged.

PR #30, opened 2015-07-02
JSON controller for similar items

rwd:
This merge would implement a JSON controller for similar items to be queried by AJAX.

The URL for a record with ID /abc/123 would be /record/abc/123/similar.json

Query parameters accepted are:
* `mltf`: field to run MLT query on, of "title", "who", "DATA_PROVIDER" or "what". If blank (default), runs all.
* `page`: page of results to retrieve
* `per_page`: items per page (default is 4)

Response looks like:

```json
{
    "page": 1,
    "per_page": 4,
    "total": 5755054,
    "documents": [
        {
            "url": "/record/09003/5487E1BD410E38C1075D8CA5D6EC3A324C2C8708.html",
            "title": "Vertrekhal van Schiphol",
            "img": {
                "alt": "Vertrekhal van Schiphol",
                "src": "http://europeanastatic.eu/api/image?uri=http%3A%2F%2Fna.memorix.nl%2Foai2%2F%3Fimage%3Dna%3Acol1%3Adat269518%3A269518.jpg%26type%3Dlarge&size=LARGE&type=IMAGE"
            }
        }
    ]
}
```


dan-nl:
is it possible to use the request header, Accept: application/json, instead of .json in the url?

rwd:
Maybe. Probably. But why?

dan-nl:
this was something that came up awhile back when someone analysed the portal api. one of their main points was that a "proper api" should use content negotiation over extension to determine the content type to deliver. i'm not sure which way the dev team decided to go ...

DavidHaskiya:
All public Europeana APIs currently have the format in the path. Let's
stick consistently to that until/if we take a considered collective
decision to switch to content negotiation across the board. But for now
let's stick with consistency and format in the path.

PR #34, opened 2015-07-03
Re: #285 restored next/prev buttons

andyjmaclean:
Follow up on pull request #28


rwd:
@andyjmaclean the Travis CI build fails with a PhantomJS on this spec

PR #54, opened 2015-07-17
Re: #281 reworked media model / bumped style

andyjmaclean:


andyjmaclean:
@rwd 
I had to rework the code that handles extra_info in the data_section function.

I could understand a refactor that you'd made to my original code so I had to rewrite it in a way I could understand.

The extra_info code works when the melkmeisje record (http://localhost:3000/record/90402/SK_A_2344.html) shows dates next to the "People" sub-heading:

Broken:
    Creator: RM0001.PEOPLE.2529 Vermeer, Johannes 

Fixed:
    Creator: RM0001.PEOPLE.2529 Vermeer, Johannes (1632 - 1632-10-31 — 1675-12 - 1675-12-15 ) 

(yes, the dates are garbled, but that's something else)

Travis / CI / hound are all happy, but coverage has dropped slightly.

PR #81, opened 2015-07-24
Feat/236 object metadata

andyjmaclean:


rwd:
@andyjmaclean what are all these merges from?

andyjmaclean:
My mistake.  I was correcting a deploy and commited and pushed without realising I was still on test branch.

To correct this I checked out the feature branch I was supposed to be on and merged test branch into it, but I used the no-ff flag, which pulled in all the comments.

Apologies.

PR #82, opened 2015-07-25
Re: #328 - serve min js

andyjmaclean:


andyjmaclean:
Use env var JS_ENTRYPOINT=/js/min/ to serve minified js.

PR #105, opened 2015-07-29
Feat/332 downloads default disabled

andyjmaclean:


andyjmaclean:
@rwd I see this has pulled in a load of comments from test again.

This cause this time was slightly different - I made the one-line change on test then checked out develop (it flagged the file with an 'M').  I then created a branch based on develop and committed that - (causing all refs to the test branch to be pulled in).  Sorry about that.

PR #264, opened 2015-11-05
Optimize views

rwd:
* Refactors named entity logic in `Portal::Show` (app/views/portal/show.rb) to prevent repetition, improve maintainability and facilitate future re-use
* If a named entity has only one value keyed by "def", just display the value, not also that key
* Stores more computed display data in an instance variable via the `mustache` helper method, to prevent multiple calls to the same template placeholder (in the Mustache templates) resulting in repeated generations of the same data
* Relies on the europeana-blacklight gem to translate pref labels


rwd:
@andyjmaclean FYI, there were made here a number of changes to the item display view class that you should be aware of.

PR #342, opened 2016-01-11
Re #203: null session when missing CSRF params

rwd:
In order to supply POST functionality when users have cookies disabled, use null session instead of exception for missing CSRF params.


rwd:
Hakiri says no.

PR #352, opened 2016-01-13
Re #786: Query for language as LANGUAGE

rwd:
Not as "what" aggregation.


rwd:
Discussion required as the output field is not part of the LANGUAGE aggregation.

See for example http://www.europeana.eu/portal/record/09404/id_oai_bc_bdsandomierz_pl_98.html, where displayed language code is "la", which has no results on the API search for LANGUAGE:la

PR #464, opened 2016-04-01
Fix opengraph tags in meta information

dennisvdvliet:
Two meta tags had the wrong name:
- fb:appid should have been fb:app_id
- og:sitename should have been og:site_name

@rwd


andyjmaclean:
My mistake - thanks!

PR #474, opened 2016-04-27
Set DJ priorities

rwd:


rwd:
Not possible to set job priorities with DJ & ActiveJob.

PR #547, opened 2016-06-27
Disable view caching by env var `DISABLE_VIEW_CACHING`

rwd:


rwd:
@andyjmaclean, you may find it useful to set this environment variable in your dev environment to prevent the caching of rendered views, without disabling caching entirely so that you still get record counts, blog feeds, etc. from the cache.

rwd:
@andyjmaclean, also, there is now a file .env.example in the app root which documents all of the available environment variables.

PR #590, opened 2016-07-19
Re: #1143 Added provisional colour model

andyjmaclean:


andyjmaclean:
The hound errors are due to a redundant variable that will no longer be redundant when I figure out how to make the functions collection_path and search_path available to app/presenters/document/web_resource_presenter.rb

PR #594, opened 2016-07-20
Static & browse page urls

rwd:
Should have .html at the end of them. Also those pages should have alternate link elements.


rwd:
Ignoring the complaints about "Missing frozen string literal comment." for existing files, for now.

PR #609, opened 2016-08-03
Re: #1307 Added grouped browse entries / bump styleguide

andyjmaclean:


andyjmaclean:
@rwd  The hound didn't email me so I never saw the bug - sorry about that

PR #628, opened 2016-09-07
Fix aliased facets

rwd:


rwd:
This should be against the release/v0.6.0 branch

PR #736, opened 2017-01-09
Re: #1793 Capitalise country names in model - revert styleguide change

andyjmaclean:


app/presenters/document/field_group_presenter.rb
```diff
@@ -100,14 +100,17 @@ def section_field_subsection(section)
       field_values.compact.map do |val|
         {}.tap do |item|
           item[:text] = val
+          if !val.nil? && section[:capitalised]
+            item[:text] = val.split.map(&:capitalize) * ' '
```
rwd: @andyjmaclean please use `item[:text] = val.titleize` instead.

PR #743, opened 2017-01-12
Feat/1772 automated facet entries

lbiedinger:


rwd:
@lbiedinger all of the API querying logic that is in app/models/facet_link_group.rb should be moved out of the model and into a job the performance of which is triggered in an after_save callback.

rwd:
@lbiedinger to apply the filters for, e.g. restricting fields to those ending "(Designer)", you will need to make use of `FacetPresenter#items_to_display`. Feel free to change its access from private to public if needed.

rwd:
@lbiedinger for a first iteration, this does the job, but could you make these changes (if you agree with them):

1. In `FacetLinkGroupGeneratorJob#set_facet_entries`, wrap the various ActiveRecord commands in a transaction to rollback gracefully if something goes wrong
2. Move the `facet_entries` and `facet_link_groups` associations from `Page` to `Page::Landing` as they are only used on the latter
3. Remove from `BrowseEntry::FacetEntry` model the `facet_field` column and instead read it from the associated facet link group to prevent duplication of data
4. In `FacetLinkGroup` combine the `after_create` and `after_save, on: :update` callbacks to just `after_save` just to DRY it up a bit
5. Change the `facet_entries` association (moved to `Page::Landing`) to: `has_many :facet_entries, through: :facet_link_groups, source: :browse_entry_facet_entries` to save writing them to the page_elements table as they are not positionable, and remove the final line in `FacetLinkGroupGeneratorJob#set_facet_entries` that writes them to page_elements. (Unless there is a need to have them in page_elements?)

PR #748, opened 2017-01-17
Query params in rel="alternate" links

rwd:


lbiedinger:
/portal/en/explore/colours.html?q=hello
still adds the q param to the alternate link.
We should reverse the logic and only ever add the parameters to search pages instead of removing it from the show page.

rwd:
@lbiedinger but if one was on /portal/en/explore/sources.html?theme=music then the alternate links should include ?theme=music, no?

Your /portal/en/explore/colours.html?q=hello example is a hypothetical that does not ever get reached from any of our internal links, right?

lbiedinger:
yes, you're right, going to merge.

PR #756, opened 2017-01-19
Cache-proof sneak peak

rwd:
Because landing page HTML is cached and the facet entry needs to be
random, so the JS should pick one from all after each page load.


app/views/collections/show.rb
```diff
@@ -82,6 +82,7 @@ def content
           hero_config: hero_config(@landing_page.hero_image),
           entry_points: facet_entry_items_grouped(@landing_page),
           preview_search_data: preview_search_data,
+          preview_search_data_present: preview_search_data.size > 0,
```
rwd: @andyjmaclean if you change this to `preview_search_data.present?` it will take care of the nil case breaking the specs (and non-browse layout landing pages) on Travis.

PR #758, opened 2017-01-24
add labeling to facet entry points, refactor facet entry point limits

lbiedinger:


app/views/concerns/facet_entry_point_displaying_view.rb
```diff
@@ -49,4 +49,11 @@ def facet_entry_field_title(facet_link_group)
     presenter = FacetPresenter.build(ff, controller)
     presenter.facet_title || facet_link_group.facet_field
   end
+
+  def facet_entry_label(entry)
```
rwd: @lbiedinger it seems quite wasteful to initialize `Europeana::Blacklight::Response::Facets::FacetField` and `FacetPresenter` for each entry when they are already grouped in this concern by facet field.

Do you think we could/should create one instance of each of those classes for each group of facet entries and extract each entry label from the one call to `FacetPresenter#display`?

Or even create a new method on `FacetPresenter` sepcifically for the content required here by facet entries?

lbiedinger: I think more specific methods for facet entries in the FacetPresenter make sense, in fact I'm thinking now it might make sense to add a FacetLinkGroupPresenter and have the things that are shared between this new presenter and the existing FacetPresenter moved to a concern under presenters.

lbiedinger: which is probably all in the presenters/concerns/facet/labelling.rb concern.

app/views/concerns/facet_entry_point_displaying_view.rb
```diff
@@ -6,47 +6,14 @@ module FacetEntryPointDisplayingView
...
-    #     title: entry.title,
-    #     url: browse_entry_path(entry, page),
-    #     image: entry.file.nil? ? nil : entry.file.url,
-    #     image_alt: nil
-    # }
-  end
-
   ##
   # @param page [Page]
   def facet_entry_items_grouped(page)
```
rwd: @lbiedinger this could be made more concise like:
```ruby
  def facet_entry_items_grouped(page)
    page.facet_link_groups.map do |facet_link_group|
      FacetLinkGroupPresenter.new(facet_link_group, controller, blacklight_config, page).display
    end
  end
```

app/presenters/facet_link_group_presenter.rb
```diff
@@ -0,0 +1,70 @@
...
+    {
+      url: browse_entry_path(entry, @page),
+      label: facet_item_label(entry.facet_value),
+      value: entry.facet_value,
+      image_url: thumb_url,
+      image_alt: nil
+    }
+  end
+
+  def respond_to_missing?(method_name, include_private = false)
```
rwd: When might `FacetLinkGroupPresenter` need to pass missing method calls to the controller? i.e. which kinds of methods?

app/views/collections/show.rb
```diff
@@ -176,5 +176,11 @@ def preview_search_data
         }
       end
     end
+
+    def facet_entry_field_title(facet_entry)
```
rwd: Could we reduce repeated execution here with something like:
```ruby
@facet_entry_field_titles ||= {}
@facet_entry_field_titles[facet_entry.facet_field] ||= begin
  ff = Europeana::Blacklight::Response::Facets::FacetField.new(facet_entry.facet_field, [])
  presenter = FacetPresenter.build(ff, controller)
  presenter.facet_title || facet_entry.facet_field
end
```
i.e. so that the instantiation of the more complex objects only occurs once for each facet field.

app/presenters/facet_link_group_presenter.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+##
+# Display logic for facet link groups(facet entry points).
+#
+# Responsible for generating data hashes for the Mustache templates from
+# [Europeana's styleguide](https://github.com/europeana/europeana-styleguide-ruby)
+# to display facet entry points on browse based landing pages.
+class FacetLinkGroupPresenter
+  include Facet::Labelling
+  include FacetsHelper
```
rwd: Does `FacetLinkGroupPresenter` ever use any of the methods from `FacetsHelper`?

app/presenters/facet_link_group_presenter.rb
```diff
@@ -0,0 +1,70 @@
...
+  include UrlHelper
+  delegate :t, to: I18n
```
rwd: Does `FacetLinkGroupPresenter` ever use `#t`?

app/presenters/facet_link_group_presenter.rb
lbiedinger: no.

app/presenters/facet_link_group_presenter.rb
lbiedinger: it does. theoretically we could leave out the delegation since we pass missing methods to the controller anyway.

app/presenters/facet_link_group_presenter.rb
lbiedinger: it checks the collection_url
also if we don't delegate t as we do above it would pass that

PR #763, opened 2017-01-31
Gallery refinements

rwd:
* Associate galleries with collections to support "themed" galleries
* Galleries have unique titles enforced not by validation but by the `Gallery` model automatically appending a number to the end of the title until it is unique
* Galleries have slugs used in their URLs, auto-generated by the stringex gem
* Gallery index and show page URLs are beneath /explore

### DIFF
--- a/app/models/gallery.rb
+++ b/app/models/gallery.rb
@@ -72,8 +72,8 @@ class Gallery < ActiveRecord::Base
   def ensure_unique_title
     i = 0
     unique_title = title
-    while !Gallery.where(title: unique_title).where.not(id: id).blank?
-      i = i + 1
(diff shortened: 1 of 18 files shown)


app/models/gallery.rb
```diff
@@ -13,10 +15,15 @@ class Gallery < ActiveRecord::Base
 
   validates :title, presence: true, length: { maximum: 60 }
   validates :description, length: { maximum: 280 }
+  validates :slug, presence: true
```
lbiedinger: there's no field in the cms to enter this, can we add a before_save/create to set the slug to an URL friendly version of the title?

rwd: @lbiedinger the stringex gem takes care of this via its `acts_as_url` helper.

PR #771, opened 2017-02-06
Handle both minimal and standard profile annotations

rwd:
Because the Annotations API is about to change its default from minimal to standard.

### DIFF
--- a/app/controllers/concerns/europeana/annotations_api_consumer.rb
+++ b/app/controllers/concerns/europeana/annotations_api_consumer.rb
@@ -6,20 +6,42 @@ module Europeana
     extend ActiveSupport::Concern
 
     def document_annotations(document)
-      search = nil
-      begin
-        search = Europeana::API.annotation.search(annotations_api_search_params(document))
-      rescue Europeana::API::Errors::ServerError
-        return nil
+      search_response = annotations_search_for_document(document)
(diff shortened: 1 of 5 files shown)


app/controllers/concerns/europeana/annotations_api_consumer.rb
```diff
@@ -6,20 +6,42 @@ module AnnotationsApiConsumer
...
-        search = Europeana::API.annotation.search(annotations_api_search_params(document))
-      rescue Europeana::API::Errors::ServerError
-        return nil
+      search_response = annotations_search_for_document(document)
+      return nil unless search_response.key?('items')
+
+      annotations_from_search_response(search_response).
+        map { |annotation| annotation_text_to_display(annotation) }.compact
+    rescue Europeana::API::Errors::ServerError
+      # @todo we may not want controller actions to fail if annotations are
```
lbiedinger: Let's do the logging as part of this already.

```
  ...
rescue Europeana::API::Errors::ServerError => server_error
  Rails.logger.error(server_error.message)
  return
end
```
could be enough for now, so at least there is some record in case we get this specific error again.

ymamakis: I also commented with Sergiu and promised that the error that he has on the annotations will be resolved as well. Thanks

PR #778, opened 2017-02-07
Re: #1867 Gallery social

andyjmaclean:


rwd:
@andyjmaclean as the `galleries_social` content is identical in these two views, could you move it into the shared `GalleryDisplayingView` concern?

PR #782, opened 2017-02-08
Topics & Categorisations

rwd:
Re: https://europeanadev.assembla.com/spaces/europeana-npc/tickets/1905

Using this requires running both `db:migrate` and `db:seed` (latter is non-destructive)

Galleries can be assigned a topic from a pre-seeded list in the CMS.

Topics are associated with galleries, via a polymorphic association on the other new model, Categorisation. The reason for this approach instead of adding a foreign key to the galleries table is that it makes it trivial in future to make other models categorisable using these topics, without adding foreign keys to their tables.

One thing that is not great yet is the way the topic input is presented in the RailsAdmin CMS for a gallery, where it appears as a nested form beneath categorisation. Any ideas on how to improve that, simplifying it to just have the topic select at the same level as the other gallery form fields are most welcome!

### DIFF
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -9,7 +9,7 @@ class Topic < ActiveRecord::Base
   accepts_nested_attributes_for :translations, allow_destroy: true
 
   acts_as_url :label, url_attribute: :slug, only_when_blank: true,
-                     allow_duplicates: false
+                      allow_duplicates: false
 
   default_scope { includes(:translations).order('topic_translations.label') }
 
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -1,5 +1,19 @@
 # frozen_string_literal: true
 class Topic < ActiveRecord::Base
+  has_many :categorisations, inverse_of: :topic, dependent: :destroy
+
+  validates :name, presence: true, uniqueness: true
+  validates :entity_uri, uniqueness: true, allow_blank: true
+
   translates :name, fallbacks_for_empty_translations: true
   accepts_nested_attributes_for :translations, allow_destroy: true
+
+  acts_as_url :name, url_attribute: :slug, only_when_blank: true,
+                     allow_duplicates: false
+
+  default_scope { includes(:translations).order('topic_translations.name') }
+
+  def to_param
+    slug
+  end
 end
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+class Topic < ActiveRecord::Base
+  translates :name, fallbacks_for_empty_translations: true
+  accepts_nested_attributes_for :translations, allow_destroy: true
+end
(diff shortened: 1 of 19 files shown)


spec/models/topic_spec.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+RSpec.describe Topic do
+  it { is_expected.to validate_presence_of(:label) }
+  it { is_expected.to validate_uniqueness_of(:label) }
+  it { is_expected.to validate_uniqueness_of(:entity_uri).allow_blank }
+  it { is_expected.to have_many(:categorisations).inverse_of(:topic).dependent(:destroy) }
+
+  it 'should translate label' do
+    expect(described_class.translated_attribute_labels).to include(:label)
```
lbiedinger: translated_attribute_names

rwd: Yes it should, and now is.

db/migrate/20170207155507_create_topics.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+class CreateTopics < ActiveRecord::Migration
+  def change
+    create_table :topics do |t|
+      t.string :slug, index: true
```
lbiedinger: why do we need a slug? topics aren't accesible via a controller, they don't need RESTfull  URIs

app/models/topic.rb
```diff
@@ -0,0 +1,19 @@
...
+class Topic < ActiveRecord::Base
+  has_many :categorisations, inverse_of: :topic, dependent: :destroy
+
+  validates :label, presence: true, uniqueness: true
+  validates :entity_uri, uniqueness: true, allow_blank: true
+
+  translates :label, fallbacks_for_empty_translations: true
+  accepts_nested_attributes_for :translations, allow_destroy: true
+
+  acts_as_url :label, url_attribute: :slug, only_when_blank: true,
```
lbiedinger: what's the point of urls for topics when we don't have a controller for them?

spec/models/topic_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  it 'should set the slug from the label' do
```
lbiedinger: see question re slugs above.

db/migrate/20170207155507_create_topics.rb
rwd: Galleries are going to be filtered by theme, where it will be better to have in the URL `?topic=fashion` than `topic=3`

app/models/topic.rb
rwd: This is a convenience method of the stringex gem which generates URL-friendly slugs from multi-lingual strings, at first only to be used in filtering galleries by topic as mentioned above.

config/initializers/rails_admin.rb
```diff
@@ -438,6 +447,25 @@
     end
   end
 
+  config.model 'Topic' do
```
lbiedinger: topics_translations.label instead of .name
also field :label

rwd: Yep. Fixed.

db/migrate/20170207155507_create_topics.rb
lbiedinger: makes sense

PR #783, opened 2017-02-08
caching for galleries

lbiedinger:


rwd:
Initial test results follow. Issues are in bold.

*Gallery foyer*

* Watching the Rails logs, caching of the foyer view HTML does occur, reducing view generation time on subsequent loads.
* **Unpublishing a gallery may cause regeneration of the cached foyer, but not always. If you have two galleries shown on the foyer, try unpublishing each in turn and reloading the foyer after each. One will not trigger regeneration of the foyer.**
* Changing image order in a gallery does always cause regeneration of the cached foyer, as does changing gallery title
* Creating and publishing a new gallery does cause regeneration of the cached foyer
* **Requesting the cached gallery foyer still results in API queries being made**

*Gallery pages*

* Rails logs shows individual gallery pages are cached
* Changing image order does cause regeneration of the cache, as does changing gallery title
* **After unpublishing a cached gallery, as a non-logged-in user I can access it by URL**
* **Requesting a cached gallery still results in API queries being made**

lbiedinger:
In the interest of not having de-published galleries on cached versions, the cache key for the index view is now updated on each edit to ANY gallery.

I was unable to reproduce accessing non-published cached pages, probably this wasn't working before as this branch was originally diverged from develop before that feature was implemented.

app/views/galleries/index.rb
```diff
@@ -38,6 +38,13 @@ def navigation
       }.reverse_merge(super)
     end
 
+    def body_cache_key
+      @body_cache_key ||= begin
+        last_galleries_edit_int = Gallery.all.order(updated_at: :desc).first.updated_at.to_i
```
rwd: You can omit the `.all` here.

app/controllers/galleries_controller.rb
```diff
@@ -1,21 +1,27 @@
 # frozen_string_literal: true
 class GalleriesController < ApplicationController
+  helper_method :documents
```
rwd: This is a clever and succinct way of preventing the API calls when there is cached HTML... _but_ API calls should really be triggered earlier than view rendering.

I'd suggest that we do something like move the body_cache_key generation for each action (and corresponding view) into the controller and have it check for the existence of that cache entry before making the API call.

lbiedinger: that was my first approach too, however it means either duplicating, somehow including or refactoring the cacheable_view.rb concern.

rwd: True, true.

A refactor would be worthwhile here I reckon, and should not be too tricky, something like: https://github.com/europeana/europeana-portal-collections/commit/4e1bea340cc3fcba630a850f6d878e90e3f9cb52

PR #785, opened 2017-02-09
Gallery image validations from EDM

rwd:
Re: https://europeanadev.assembla.com/spaces/europeana-npc/tickets/1906

### DIFF
--- a/spec/models/gallery_spec.rb
+++ b/spec/models/gallery_spec.rb
@@ -1,5 +1,50 @@
 # frozen_string_literal: true
 RSpec.describe Gallery do
+  before do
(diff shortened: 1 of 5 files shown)


spec/models/gallery_spec.rb
```diff
@@ -1,5 +1,24 @@
 # frozen_string_literal: true
 RSpec.describe Gallery do
+  before do
```
lbiedinger: let's do some validation tests by adding a context where the api_response is one with a URL without type image.
(expected to not be valid)

rwd: Indeed. See following commit.

PR #786, opened 2017-02-09
Gallery foyer pagination

rwd:
A default of 24 galleries are shown per page. This can be overriden via the `per_page` URL param, which will facilitate testing.

### DIFF
--- a/app/controllers/galleries_controller.rb
+++ b/app/controllers/galleries_controller.rb
@@ -37,10 +37,10 @@ class GalleriesController < ApplicationController
   end
 
   def gallery_page
-    params[:page].to_i || 1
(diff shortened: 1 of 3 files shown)


app/controllers/galleries_controller.rb
```diff
@@ -35,4 +35,12 @@ def homepage_hero_image
     landing_page = Page::Landing.find_by_slug('')
     landing_page.nil? ? nil : landing_page.hero_image
   end
+
+  def gallery_page
+    (params[:page] || 1).to_i
+  end
+
+  def gallery_per
+    (params[:per_page] || 24).to_i
```
lbiedinger: there's currently no way to set this through the UI, right?

PR #788, opened 2017-02-10
add published_at and published_by to galleries

lbiedinger:
this adds two columns to the galleries table, these are used for tracking the user and date of the first publication of a gallery.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 ##
 # User model
 class User < ActiveRecord::Base
@@ -16,6 +17,8 @@ class User < ActiveRecord::Base
 
   delegate :can?, :cannot?, to: :ability
   delegate :role_enum, to: :class
+
+  has_many :galleries, inverse_of: :publisher
 
   class << self
     def role_enum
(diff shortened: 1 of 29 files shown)


db/schema.rb
```diff
@@ -291,6 +303,26 @@
 
   add_index "searches", ["user_id"], name: "index_searches_on_user_id", using: :btree
 
+  create_table "topic_translations", force: :cascade do |t|
```
rwd: @lbiedinger The schema committed here includes work from another branch for the topics and categorisations but which has not yet been merged. Could you remove these?

lbiedinger:
@rwd, yes, cleaned up the schema.rb, thanks for spotting!

app/models/user.rb
```diff
@@ -17,6 +18,8 @@ class User < ActiveRecord::Base
   delegate :can?, :cannot?, to: :ability
   delegate :role_enum, to: :class
 
+  has_many :galleries, inverse_of: :publisher
```
rwd: I think you need to state the name of the foreign key column here as otherwise this fails: `Gallery.where.not(published_by: nil).first.publisher.galleries`

config/initializers/rails_admin.rb
```diff
@@ -189,11 +189,15 @@
...
       field :state
+      field :publisher
+      field :published_at
     end
     show do
       field :title
       field :description
       field :state
+      field :publisher
+      field :published_by
```
rwd: This should be `published_at`

PR #791, opened 2017-02-14
Filtering of galleries

lbiedinger:
allows filtering of galleries via topics
note: the url parameter is theme as that is what it was called previously

### DIFF
--- a/app/controllers/galleries_controller.rb
+++ b/app/controllers/galleries_controller.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 class GalleriesController < ApplicationController
   def index
-    @galleries = Gallery.includes(:images).published.order(published_at: :desc).page(gallery_page).per(gallery_per).with_topic(gallery_topic)
+    @galleries = Gallery.includes(:images).published.order(published_at: :desc).
+                 page(gallery_page).per(gallery_per).with_topic(gallery_topic)
     @selected_topic = gallery_topic
     @documents = search_api_for_image_metadata(gallery_images_for_foyer(@galleries))
     @hero_image = homepage_hero_image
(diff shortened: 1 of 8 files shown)


app/models/topic.rb
```diff
@@ -13,6 +13,10 @@ class Topic < ActiveRecord::Base
 
   default_scope { includes(:translations) }
 
+  scope :with_galleries, -> do
```
lbiedinger: https://github.com/bbatsov/rubocop/issues/1520

app/models/gallery.rb
```diff
@@ -5,6 +5,10 @@ class Gallery < ActiveRecord::Base
   include HasPublicationStates
   include IsCategorisable
 
+  scope :with_topic, ->(topic_slug) do
+    topic_slug == 'all' ? all : joins(:categorisations).joins(:topics).where("\"topics\".\"slug\" like '#{topic_slug}'")
```
rwd: 1. This is unsafe as it could be exploited for SQL injection from the theme param.
2. This returns duplicates if a gallery has more than one categorisation.

To fix both these issues, change it to: `joins(:categorisations, :topics).where(topics: { slug: topic_slug }).distinct`

app/controllers/galleries_controller.rb
```diff
@@ -43,4 +44,8 @@ def gallery_page
   def gallery_per
     (params[:per_page] || 24).to_i
   end
+
+  def gallery_topic
+    (params[:theme] || 'all')
```
rwd: You don't need parentheses here.

app/models/topic.rb
```diff
@@ -13,6 +13,10 @@ class Topic < ActiveRecord::Base
...
+    includes(:translations).joins(:categorisations).joins('INNER JOIN galleries on (categorisations.categorisable_id = galleries.id)')
```
rwd: 1. You can omit `includes(:translations)` here as it's inherited from the default scope.
2. You also need a join condition of `categorisations.categorisable_type="Gallery"`

PR #793, opened 2017-02-14
Bump europeana-logging gem version

rwd:
To include session ID in payload


rwd:
Going to recreate this.

PR #803, opened 2017-02-20
add newsletter_url to landing pages

lbiedinger:
This adds a newsletter_url column to the pages table and uses it instead of the one from the collections table.
Removal of the collections table's column should happen separately, and after these changes have been released.

### DIFF
--- a/app/views/collections/show.rb
+++ b/app/views/collections/show.rb
@@ -106,10 +106,10 @@ module Collections
     private
 
     def newsletter_content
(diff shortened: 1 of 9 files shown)


lbiedinger:
1) yes, sorry, forgot to git add, it's here now
2) I've added an initial seed value for the fashion landing page using some of the values from our current test environment, this should be updated, once there's a more final version available.
3) Removed

PR #807, opened 2017-02-24
initial work on feed modularization

lbiedinger:
Stops having a single tumblr feed hardcoded and instead introduces a new Feed model which can be managed through the CMS. Feeds are then added to landing pages through the CMS. (the association is on pages though to keep it more generic) 
Some of the responsibility for feed data retrieval is moved out of the collection helper and into the feed helper to make it more generic.
the SCHEDULE_FEED_TUMBLR and SCHEDULE_FEED_BLOG config entries are changed to SCHEDULE_FEED_CUSTOM, and decides when all Feed type feeds are harvested.
Blog feeds are still automatically added for display in collections, without using the new Feed class.
The seeds have been updated to add the previously hardcoded tumblr feed to the DB.
The tumblr json that was accessible at portal/en/collection/name/tumblr has been removed, in favour of increasing the default per page for the tumblr feed to 20(which is how many entries we're caching). @rwd  should this be put back? it doesn't seem to be used anywhere else by us though.


db/seeds/landing_pages.rb
```diff
@@ -325,6 +325,9 @@ def find_or_download_styleguide_image(path)
           settings_category: 'exhibition',
           media_object: find_or_download_styleguide_image('sample/thumb-machines.jpg')
         )
+      ],
+      feeds: [
+          Feed.find_by_slug('fashion_tumblr')
```
rwd: Shouldn't the feed slug here be 'fashion-tumblr'?

db/seeds/feeds.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+unless Feed.find_by_slug('fashion-tumblr').present?
+  Feed.create!(
+    name: 'Fashion Tumblr',
+    url: 'http://europeanafashion.tumblr.com/rss'
+  )
+
+  # if the landing page is already present we need to add our feed to it here
```
rwd: This seems to duplicate the seeding in db/seeds/landing_pages.rb. I think it belongs in a migration rather than the seeds.

spec/fixtures/feeds.yml
```diff
@@ -0,0 +1,4 @@
+fashion_tumblr:
+  name: Fashion Tumblr
+  slug: fashion_tumblr
```
rwd: If auto-generated the slug would be "fashion-tumblr", so should probably be that in the fixture too.

app/helpers/feed_helper.rb
```diff
@@ -14,4 +15,75 @@ def cached_feed(url)
...
+    return nil unless feed
+
+    items = feed_items_for(feed, options)
+
+    return nil if items.blank?
+
+    paginated_items = Kaminari.paginate_array(items).page(page).per(per_page)
+    {
+      title: 'Tumblr',
+      tumblr_url: Cache::FeedJob::URLS[:custom][feed.slug.to_sym].sub('/rss', ''),
```
rwd: `Cache::FeedJob::URLS[:custom][feed.slug.to_sym]` can be replaced with `feed.url`

app/jobs/cache/feed_job.rb
```diff
@@ -1,19 +1,25 @@
...
       },
       exhibitions: %i(de en).each_with_object({}) do |locale, hash|
         hash[locale] = (ENV['EXHIBITIONS_HOST'] || 'http://www.europeana.eu') + "/portal/#{locale}/exhibitions/feed.xml"
-      end,
-      tumblr: {
-        fashion: 'http://europeanafashion.tumblr.com/rss'
-      }
-    }
+      end
+    }.merge custom_feeds
```
rwd: This will not work as intended. Once `Cache::FeedJob::URLS` has been set, it will not be altered to reflect any new/modified/deleted feed URLs.

I suggest removing the `.custom_feeds` method and instead just calling `Feed.all` where you have called `Cache::FeedJob::URLS[:custom]`.

app/helpers/feed_helper.rb
```diff
@@ -14,4 +15,75 @@ def cached_feed(url)
...
+        date: I18n.l(item.published, format: :short),
+        excerpt: {
+          short: ActionController::Base.helpers.strip_tags(CGI.unescapeHTML(item.summary))
+        },
+        type: detect_feed_type_from_url(url)
+      }
+    end
+  end
+
+  def detect_feed_type_from_url(feed_url)
```
rwd: I would be tempted to extract the `#tumblr?` etc methods from `Link::SocialMedia` into a model concern and also include that in `Feed`, along with `Link#url_in_domain?`, and reuse that here.

app/models/feed.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+class Feed < ActiveRecord::Base
+  has_and_belongs_to_many :pages
+
+  validates :name, presence: true, uniqueness: true
+  validates :url, presence: true, uniqueness: true
+
+  acts_as_url :name, url_attribute: :slug, only_when_blank: true, allow_duplicates: false
```
rwd: Do you anticipate us needing a URL slug for `Feed`?

app/helpers/feed_helper.rb
```diff
@@ -14,4 +15,75 @@ def cached_feed(url)
...
+      {
+        url: CGI.unescapeHTML(item.url),
+        img: {
+          src: feed_entry_thumbnail_url(item),
+          alt: item.title
+        },
+        title: item.title,
+        date: I18n.l(item.published, format: :short),
+        excerpt: {
+          short: ActionController::Base.helpers.strip_tags(CGI.unescapeHTML(item.summary))
```
rwd: This is already a helper method, so `ActionController::Base.helpers.` should not be needed.

rwd:
@lbiedinger as each blog feed is used on a landing page, we might as well move them into the feeds table to have fewer hard-coded.

rwd:
@lbiedinger if we are switching to not have more feed items loaded in by AJAX, then it's fine to have removed the Tumblr JSON.

app/models/feed.rb
lbiedinger: I had written a Feeds controller to fill in for the Collections tumblr AJAX. I didn't commit it since I later removed that functionality altogether since tumblr currently is the feed with the most items(20) and that can be loaded at once. 
If we start caching longer feeds(since we allow any custom feed), re-introducing the AJAX through a feeds controller might make sense, however I didn't want to spend more time fleshing this out without knowing what the requirements/or even which "longer feed" we want to display.

db/seeds/feeds.rb
lbiedinger: I was trying to catch all scenarios in terms of (re)running combinations of migration and seed tasks. Considering seed should only be ran once, You're right though, just moving this to the migration is the best option.

app/models/feed.rb
rwd: Fair enough. Let's leave slug there for now, it's good to have in case we need it, which there is a fair chance of, and if not we can easily drop it later.

lbiedinger:
@rwd All the blog feeds are now also moved to the Feed model, this means the scheduled jobs for the blog are obsolete and I've removed the SCHEDULE_FEED_BLOGS var from the .env file.

app/helpers/feed_helper.rb
```diff
@@ -14,4 +15,61 @@ def cached_feed(url)
   def feed_entry_thumbnail_url(entry)
     FeedEntryImage.new(entry).thumbnail_url
   end
+
+  def tumblr_feed_content(landing_page, options = {})
```
rwd: Why do we have this method? It's only called when the landing page layout is "default", but we only have a Tumblr feed for the fashion collection landing page which uses the "browse" layout.

app/models/feed.rb
```diff
@@ -0,0 +1,18 @@
...
+  include HasExternalUrls
+
+  has_and_belongs_to_many :pages
+
+  validates :name, presence: true, uniqueness: true
+  validates :url, presence: true, uniqueness: true
+
+  acts_as_url :name, url_attribute: :slug, only_when_blank: true, allow_duplicates: false
+
+  scope :tumblr, -> { where('url LIKE (?)', '%tumblr.com%') }
```
rwd: I'm not too keen on this duplication of the URL patterns already present in the `HasExternalUrls` concern.

Where you have used `landing_page.feeds.tumblr.first` and `landing_page.feeds.blog.first`, could we just use `landing_page.feeds.detect(&:tumblr?)` and `landing_page.feeds.detect(&:europeana_blog?)` instead, and remove these scopes?

app/views/concerns/navigable_view.rb
```diff
@@ -131,7 +131,7 @@ def navigation_global_primary_nav_exhibitions_submenu_items
   end
 
   def navigation_global_primary_nav_blog_submenu_items
-    feed_items = feed_entry_nav_items(Cache::FeedJob::URLS[:blog][:all], 6)
+    feed_items = feed_entry_nav_items(Feed.find_by_slug('all-blog').url, 6)
```
rwd: It is possible that this feed will not be found, e.g. if it gets deleted via the CMS. This code needs to handle the case when no record is found.

app/models/feed.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+class Feed < ActiveRecord::Base
```
rwd: Could we add a method to `Feed` to return the URL of the HTML version of it, like:

```ruby
def html_url
  if tumblr?
    url.sub('/rss', '')
  elsif europeana_blog?
    url.sub('/feed', '')
  else
    url
  end
end
```

That could then be used in`Home::Index#content` (view), `FeedHelper#tumblr_feed_content`, `NavigableView#navigation_global_primary_nav_blog_submenu_items` (instead of hard-coded 'http://blog.europeana.eu/'), and anywhere else we in future need such URLs.

lib/clock.rb
```diff
@@ -5,21 +5,15 @@
...
       Cache::FeedJob.perform_later(url)
     end
   end
 
-  every(1.day, 'cache.feed.tumblr', at: ENV['SCHEDULE_FEED_TUMBLR']) do
-    Cache::FeedJob::URLS[:tumblr].values.each do |url|
-      Cache::Feed::BlogJob.perform_later(url)
+  every(1.day, 'cache.feed.custom', at: ENV['SCHEDULE_FEED_CUSTOM']) do
+    Feed.all.each do |feed|
+      Cache::Feed::BlogJob.perform_later(feed.url)
```
rwd: We probably want to rename `Cache::Feed::BlogJob` to not mention blogs. Or just move its image downloading logic into `Cache::FeedJob` but make it optional by argument to `#perform` so that exhibitions feeds can elect to skip it.

app/models/feed.rb
rwd: Also, we probably want an ActiveModel callback to queue the feed caching job every time that a feed is created or updated.

app/helpers/feed_helper.rb
lbiedinger: in production the fashion collection still uses the default layout.

rwd: Of course it does! Thanks!

app/models/feed.rb
lbiedinger: I'm not entirely convinced of how useful this method is for other feeds that may not have human readable versions.
Or where the human readable version can't be easily derived from the RSS URL. At some point it may make sense to add the html_url for each feed to the DB.
In any case, for the moment this is indeed quite useful for our purposes and the feeds we're currently dealing with, so I'll add this.

rwd: Agreed. We will revisit and refactor this if/when it becomes inadequate.

PR #815, opened 2017-02-28
Add respond_to to galleries controller actions

rwd:
So that appending .json does not render the source code...


rwd:
@lbiedinger this is in response to an undesired behaviour noticed by @andyjmaclean where "you get the source code on the screen if you append .json to the galleries url: http://localhost:3000/portal/en/explore/galleries.json"

PR #819, opened 2017-03-01
Re: #1956 added missing urls

andyjmaclean:


andyjmaclean:
@lbiedinger or @rwd I use the collection name to derive a link to the collection.  Can't get the provider image to come through locally - even though I'm including an item that has a data provider with an assigned image (like this, which I think has been given an image on test though the server is down and I can't check http://test-npc.eanadev.org/portal/en/record/2059516/data_foodanddrink_O99682.html)

app/views/galleries/show.rb
```diff
@@ -44,6 +44,14 @@ def gallery_items_content
       mustache[:gallery_items_content] ||= @gallery.images.map { |image| gallery_item_content(image) }
     end
 
+    def data_provider_logo_url(image)
+      presenter = presenter_for_gallery_image(image)
+      data_provider_name = presenter.field_value('aggregations.edmDataProvider')
```
lbiedinger: use 'dataProvider' instead of 'aggregations.edmDataProvider'

app/views/galleries/show.rb
```diff
@@ -52,7 +60,10 @@ def gallery_item_content(image)
         creator: presenter.field_value('dataProvider'),
         thumb_url: gallery_image_thumbnail(image),
         full_url: presenter.field_value('aggregations.edmIsShownBy'),
-        rights: presenter.simple_rights_label_data
+        rights: presenter.simple_rights_label_data,
+        url_item: image.portal_url,
+        url_collection: search_path(q: "europeana_collectionName:#{presenter.field_value('europeanaCollectionName')}"),
+        institution_logo: data_provider_logo_url(image)
```
lbiedinger: just pass the presenter, and you won't need to look up what the presenter is in data_provider_logo_url

PR #826, opened 2017-03-06
Re: #1959 Take total available from ajax result

andyjmaclean:


andyjmaclean:
Updated styleguide with fix

PR #829, opened 2017-03-08
Re: #1978 connect collections and landing pages

lbiedinger:
These changes link collections and landing pages. This refactor is mainly in preparation for adding more authorization logic, so it also makes various adjustments to the CMS, most of which restrict the ability to change collection keys, or associations between collections and landing pages. This is in order to only allow these values to be set on creation of new collections/landingpages(which only administrators will be able to do)

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -4,7 +4,7 @@ class Collection < ActiveRecord::Base
   include HasSettingsAttribute
 
   has_and_belongs_to_many :browse_entries
-  has_one :page_landing, class_name:'Page::Landing', dependent: :destroy
+  has_one :page_landing, class_name: 'Page::Landing', dependent: :destroy
 
   has_paper_trail
 
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -4,7 +4,7 @@ class Collection < ActiveRecord::Base
   include HasSettingsAttribute
 
   has_and_belongs_to_many :browse_entries
-  has_one :page_landing, class_name: 'Page::Landing', dependent: :destroy
+  has_one :landing_page, class_name: 'Page::Landing', dependent: :destroy
 
   has_paper_trail
 
@@ -43,10 +43,6 @@ class Collection < ActiveRecord::Base
 
   def has_landing_page?
     landing_page.present?
-  end
-
-  def landing_page
-    @landing_page ||= page_landing
   end
 
   def landing_page_title
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -4,6 +4,7 @@ class Collection < ActiveRecord::Base
   include HasSettingsAttribute
 
   has_and_belongs_to_many :browse_entries
+  has_one :page_landing, :class_name => 'Page::Landing', dependent: :destroy
 
   has_paper_trail
 
@@ -45,11 +46,7 @@ class Collection < ActiveRecord::Base
   end
 
   def landing_page
-    @landing_page ||= Page::Landing.find_by_slug(landing_page_slug)
-  end
-
-  def landing_page_slug
-    key == 'all' ? '' : "collections/#{key}"
+    @landing_page ||= page_landing
   end
 
   def landing_page_title
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -4,7 +4,7 @@ class Collection < ActiveRecord::Base
   include HasSettingsAttribute
 
   has_and_belongs_to_many :browse_entries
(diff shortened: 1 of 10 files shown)


config/initializers/rails_admin.rb
```diff
@@ -433,7 +439,14 @@
       field :feeds
     end
     edit do
-      field :slug
+      field :collection do
+        visible do
+          true unless bindings[:object].persisted?
+        end
+        associated_collection_scope do
+          proc { |_scope| Collection.published.includes(:page_landing).where(pages: { collection_id: nil }) }
```
rwd: I think it's a bit tidier to do `proc { |scope| scope.published.includes(:page_landing).where(pages: { collection_id: nil }) }`

db/migrate/20170307114233_add_collection_to_pages.rb
```diff
@@ -0,0 +1,19 @@
...
+  def change
+    add_reference :pages, :collection, index: true
+    add_foreign_key :pages, :collections
+
+    reversible do |change|
+      change.up do
+        Page::Landing.all.each do |landing_page|
+          key = landing_page.slug.split('/').last
+          key = key.blank? ? 'all' : key
+          collection = Collection.find_by_key(key)
```
rwd: As it is until now possible to create a landing page without a corresponding collection, could we, to be on the safe side, add a check whether the collection is found here, around the next two statements?

spec/fixtures/pages.yml
```diff
@@ -43,11 +47,11 @@ fashion_collection:
 
 grid_layout_collection:
   slug: 'collections/grid_layout'
+  collection: grid_layout
   state: 1
   type: 'Page::Landing'
 
 draft_landing_page:
```
rwd: Is there any point in keeping this fixture as landing pages need to be associated with collections now?

app/models/collection.rb
```diff
@@ -4,6 +4,7 @@ class Collection < ActiveRecord::Base
   include HasSettingsAttribute
 
   has_and_belongs_to_many :browse_entries
+  has_one :page_landing, class_name: 'Page::Landing', dependent: :destroy
```
rwd: I'm tempted to call this association `landing_page`, alleviating the need for the method of that name below, and which is more readable.

spec/models/page/landing_spec.rb
```diff
@@ -15,9 +16,44 @@
...
+  it { is_expected.to validate_presence_of(:collection) }
+  it { is_expected.to validate_uniqueness_of(:collection) }
 
   describe 'modules' do
     subject { described_class }
     it { is_expected.to include(PaperTrail::Model::InstanceMethods) }
   end
+
+  describe 'creation' do
+    subject { described_class.new }
```
rwd: I think this is always the default if no subject is declared.

db/migrate/20170307114233_add_collection_to_pages.rb
lbiedinger: sure, I'll not do anything with the landing page in that case, however it'll not be a valid instance in accordance with the validations on the new landing page model. If there are landing pages like this we will most likely have to do a manual cleanup anyway.

spec/fixtures/pages.yml
lbiedinger: it's being used in the ability specs, I'll assign a collection to it and keep it.
The only reason i didn't was because I was reusing it for the 'has_landing_page? when there is NO landing page' spec, but i'll use something else there.

app/models/page/landing.rb
```diff
@@ -17,9 +18,12 @@ class Page::Landing < Page
   has_settings :layout_type
 
   validates :settings_layout_type, inclusion: { in: :settings_layout_type_enum }
+  validates :collection, presence: true, uniqueness: true
 
   delegate :settings_layout_type_enum, to: :class
 
+  before_create :set_slug if :collection
```
rwd: Could you do this using an :if option to the callback instead of having the condition inline? Per http://guides.rubyonrails.org/v4.2/active_record_callbacks.html#conditional-callbacks

config/initializers/rails_admin.rb
```diff
@@ -433,7 +439,16 @@
...
     end
     edit do
-      field :slug
+      field :collection do
+        visible do
+          true unless bindings[:object].persisted?
+        end
+        associated_collection_scope do
+          proc do |_scope|
+            Collection.published.includes(:landing_page).where(pages: { collection_id: nil })
```
rwd: `proc { |scope| scope.published # ... }` ?

PR #832, opened 2017-03-08
cache hero_image_url for galleries head meta

lbiedinger:
This adds the hero_image_url for gallery show pages to the cache so it's also available when we don't make an API request because we have cached the page body.

I guess there is a small chance that the image url which is written to the cache expires a few milliseconds before the page's cached body expires, in which case it potentially could lead to the og metadata being left blank for a request during that time. However the likelihood of this is very remote and the consequences not really significant.

### DIFF
--- a/app/controllers/galleries_controller.rb
+++ b/app/controllers/galleries_controller.rb
@@ -24,7 +24,13 @@ class GalleriesController < ApplicationController
     images = @gallery.images
 
     @body_cache_key = 'explore/' + @gallery.cache_key
-    @documents = search_api_for_image_metadata(images) unless body_cached?
+
+    if body_cached?
+      @hero_image_url = gallery_hero_image_from_cache
+    else
+      @documents = search_api_for_image_metadata(images)
+      set_gallery_hero_image_cache(images.first)
+    end
 
     respond_to do |format|
       format.html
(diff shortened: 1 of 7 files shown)


app/controllers/galleries_controller.rb
```diff
@@ -48,6 +54,16 @@ def search_api_for_image_metadata(images)
...
   end
 
+  def gallery_hero_image_from_cache
+    Rails.cache.fetch(cache_key(@body_cache_key) + '/hero_image_url')
+  end
+
+  def set_gallery_hero_image_cache(image)
+    image_document = @documents.detect { |document| document.fetch(:id, nil) == image.europeana_record_id }
+    @hero_image_url = image_document['edmIsShownBy'].first
+    Rails.cache.write(cache_key(@body_cache_key) + '/hero_image_url', @hero_image_url, expires_in: 24.hours)
```
rwd: @lbiedinger should we cache this for 25 hours? The cached HTML would always then expire first and the cached hero image URL get overwritten when the former is next generated.

lbiedinger: we can do something like that, but even a minute should suffice so how about: 
`...expires_in: 24.hours + 1.minute)`

rwd: Go for it. If the HTML takes more than a minute to generate, we have bigger problems...

PR #833, opened 2017-03-08
Blog listing display

rwd:
* Extracts view logic for pagination into a PaginatedView concern, also used by main item search results and galleries listing
* Changes the Pro URL env var to be the base URL for the entire site, not that for the JSON API
* Implements most display elements for blog listing page, only missing read time (to come later)

### DIFF
--- a/app/controllers/blog_posts_controller.rb
+++ b/app/controllers/blog_posts_controller.rb
@@ -6,6 +6,8 @@
 # @todo Exception handling when `JsonApiClient` requests fail
 # @todo Extract pagination into a controller concern
 class BlogPostsController < ApplicationController
+  include HomepageHeroImage
+
   def index
     @blog_posts = Pro::BlogPost.includes(:network).page(blog_posts_page).per(6).all
     @hero_image = homepage_hero_image
@@ -15,10 +17,5 @@ class BlogPostsController < ApplicationController
 
   def blog_posts_page
     (params[:page] || 1).to_i
-  end
-
-  def homepage_hero_image
-    landing_page = Page::Landing.find_by_slug('')
-    landing_page.nil? ? nil : landing_page.hero_image
   end
 end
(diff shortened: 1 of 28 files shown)


app/controllers/blog_posts_controller.rb
```diff
@@ -8,11 +8,17 @@
...
   end
 
   protected
 
   def blog_posts_page
     (params[:page] || 1).to_i
   end
+
+  def homepage_hero_image
+    landing_page = Page::Landing.find_by_slug('')
```
rwd: I will use a class method for `Page::Landing.home` so that it only returns the one instance, not a scope which would return a collection with only one member.

I will put `homepage_hero_image` into a concern instead of a helper because db calls should be made before the view is rendered, and concerns better isolate to the contexts that need the methods than helpers.

spec/models/pro_spec.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+RSpec.describe Pro do
+  describe '.site' do
+    it 'should default to "http://pro.europeana.eu"' do
+      expect(described_class.site).to eq('http://pro.europeana.eu')
```
rwd: I don't think that would work as by the time the RSpec examples are run, that call to ENV has already occurred. I have reworked the specs to not fail when EUROPEANA_PRO_URL is set in the test env.

app/views/portal/index.rb
```diff
@@ -4,6 +4,7 @@ module Portal
   class Index < ApplicationView
     include SearchableView
     include HeroImageDisplayingView
+    include PaginatedView
```
rwd: I have added a few specs for this, split over shared examples and unit-level tests of the concern itself, which approach we could use as the basis for a general strategy for testing views.

lbiedinger: great!

PR #835, opened 2017-03-10
authorization for editors

lbiedinger:


app/models/concerns/is_permissionable.rb
```diff
@@ -0,0 +1,36 @@
...
+      joins(:permissions).where(permissions: { user_id: user })
+    end
+
+    # We need to specify this so that for STI models like Page::Landing
+    # the permissionable_type of their base class is used.
+    # This is important, as otherwise the permission won't be
+    # removable when the delete method is called on user.permissionables,
+    # as that "has_many through:" relationship will use reflection and by default
+    # this will arrive at the base class only.
+    # In order to differntiate between subclasses that are permissionable,
```
rwd: "differentiate"?

config/initializers/rails_admin.rb
```diff
@@ -152,8 +152,16 @@
       end
     end
     edit do
-      field :uri
-      field :name
+      field :uri do
+        read_only do
+          true unless bindings[:view]._current_user.role == 'admin'
```
rwd: If we really want per-attribute permissions, then they should be declared in `Ability`, with custom methods if necessary. not here.

app/models/user.rb
```diff
@@ -24,6 +34,23 @@ class << self
     def role_enum
       %w(admin editor user)
     end
+
+    def permissionable_landing_page_ids_enum
+      puts Page::Landing.all.inspect
```
lbiedinger: yes! Sorry, I really got to stop leaving these in!

app/models/browse_entry.rb
```diff
@@ -23,6 +24,7 @@ class BrowseEntry < ActiveRecord::Base
   accepts_nested_attributes_for :translations, allow_destroy: true
   default_scope { includes(:translations) }
 
+  before_create :set_editor_permissions
```
lbiedinger: yeah, that's exactly the reason I didn't put it in to the concern. And I guess the fact that we don't want it on landing pages made me consider future Classes that shouldn't have it, but let's deal with that as we need to, we can always move it out if required.

config/initializers/rails_admin.rb
```diff
@@ -509,6 +517,25 @@
       field :password
       field :password_confirmation
       field :role
+      group :permissions do
```
lbiedinger: probably we want it in three places, on the users, on the permissionables and it's own section. On the other hand, I don't think the permissions page would be extremely useful as it wouldn't have much context/order, as in a way it would just be a list of all users and all permissionable models thrown together.

app/models/concerns/is_permissionable.rb
```diff
@@ -0,0 +1,36 @@
...
+    # a lambda with a where clause on the type can be defined on users.
+    def permissionable_type=(class_name)
```
lbiedinger: no, will move them out

config/initializers/rails_admin.rb
lbiedinger: there are a few ways descirbed to do this here: http://stackoverflow.com/questions/5921591/cancan-limiting-a-users-ability-to-set-certain-model-attributes-based-on-their
None of them are really very nice. Using cancan2 which never came out of beta, modifying ActiveRecord::Base, or just hiding the field like what was happening here. Since not having these fields editable was not a requirement on the ticket, I'll just leave them. Should this be an issue, we can reconsider the best approach.
In which case I would also suggest the option of just adding the DataProviderImage class to the cms and giving editors rights on that instead, if that's all they should be updating.

PR #837, opened 2017-03-13
Blog post page

rwd:
* The blog listing display now has links to display individual items
* Display logic for blog posts has been extracted into a presenter class used by both index and show actions


app/controllers/blog_posts_controller.rb
```diff
@@ -11,12 +11,20 @@ class BlogPostsController < ApplicationController
...
     @hero_image = homepage_hero_image
   end
 
+  def show
+    @blog_post = scope.where(slug: params[:slug]).first
+  end
+
   protected
 
+  def scope
```
lbiedinger: is there a more meaningful name for this method than just "scope"?

rwd: Yes: "pro_blog_posts". Thanks!

lbiedinger: pro_blog_posts

PR #841, opened 2017-03-15
Bug/1977 1986 https rights statements

lbiedinger:
Allow both http:// and https:// for rights statement urls
basically this PR introduces the s? wildcard to all the rights urls that are used. It also makes the conversion of the pattern string more versatile by removing any wildcard elements be it single character [x?] or multi character patterns [(abcdx)?] with [*] so the pattern can be used as an api param.
This PR also adds a reverse lookup from the params, which strips wildcards, so that the facets can be identified properly on search pages.

### DIFF
--- a/app/models/edm/rights.rb
+++ b/app/models/edm/rights.rb
@@ -17,7 +17,7 @@ module EDM
       end
 
       def for_api_query(value)
-        registry.detect { |rights| value.match(rights.pattern)}
+        registry.detect { |rights| value.match(rights.pattern) }
       end
 
       def from_api_query(value)
--- a/app/models/edm/rights.rb
+++ b/app/models/edm/rights.rb
@@ -16,10 +16,6 @@ module EDM
         registry.detect { |rights| string.match(rights.pattern) }
       end
 
-      def for_api_query(value)
-        registry.detect { |rights| value.match(rights.pattern) }
-      end
-
       def from_api_query(value)
         simple_value = value.to_s.tr('?*', '')
         registry.detect { |rights| simple_value.match(rights.pattern) }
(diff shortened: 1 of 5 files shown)


lbiedinger:
@rwd 

For your query the API returns the facet with a value of:
`{"name":"REUSABILITY","fields":[{"label":"permission","count":1}]}`

Which is also something that happens on a query like the one which is created by the facet now: 
/portal/en/search?f[REUSABILITY][]=open&f[RIGHTS][]=http*%3A%2F%2Fcreativecommons.org%2Flicenses%2Fby%2F*&per_page=12&q=&view=grid

`{"name":"REUSABILITY","fields":[{"label":"open","count":3270694},{"label":"permission","count":8213}]}`

This seems like an issue at the API level. It seems like all https rights are given the "permission" reusability.

http://www.europeana.eu/api/v2/search.json?query=provider_aggregation_edm_rights:https*&profile=facets

app/models/edm/rights.rb
```diff
@@ -17,12 +17,17 @@ def normalise(string)
       end
 
       def for_api_query(value)
```
rwd: `#for_api_query` is no longer used anywhere in the app. If that is intentional, please remove it.

app/models/edm/rights.rb
```diff
@@ -17,12 +17,17 @@ def normalise(string)
...
+
+      def from_api_query(value)
+        simple_value = value.to_s.tr('?*', '')
+        registry.detect { |rights| simple_value.match(rights.pattern) }
       end
     end
 
     def api_query
-      super || pattern + '*'
+      super || pattern.gsub(/\(.*\)\?|.\?/, '*') + '*'
```
rwd: As api_query would now always be derived and never declared explicitly in rights.yml, the call to `super` can be removed.

PR #847, opened 2017-03-21
Events from Pro via JSON API

rwd:
Initial implementation of events retrieval from Pro site's JSON API.

* Views show some basic data as a POC; full population pending template implementation
* Pagination of JSON API result sets extracted into a new view concern
* Pagination in controllers extracted into a controller concern, but not applied to `PortalController` as that has multiple actions paginating different result sets, plus Blacklight handles pagination for `#index`

### DIFF
--- a/app/models/pro/event.rb
+++ b/app/models/pro/event.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+module Pro
+  ##
+  # Events from Pro JSON API.
+  class Event < Base
+  end
+end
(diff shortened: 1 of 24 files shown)


app/models/pro/event.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+module Pro
+  ##
+  # Events from Pro JSON API.
+  class Event < Base
+  end
```
lbiedinger: we don't know the "table_name" yet right?

rwd: `JSON::API` correctly guesses that the "table name" for this model is "events" so no need to set it explicitly.

lbiedinger: cool, I hadn't realised this was all in place already, but yes it's all working. There's going to probably be some refactor for when we get the design all implemented, but this all looks good.

PR #848, opened 2017-03-21
add galleries to top nav

lbiedinger:


app/views/concerns/navigable_view.rb
```diff
@@ -138,6 +140,23 @@ def navigation_global_primary_nav_blog_submenu_items
...
+        items: navigation_global_primary_nav_galleries_submenu_items
+      }
+    }
+  end
+
+  def navigation_global_primary_nav_galleries_submenu_items
+    Gallery.published.order(published_at: :desc).map do |gallery|
+      link_item(gallery.title, gallery_path(gallery),
+                is_current: current_page?(gallery_path(gallery)))
+    end << link_item(t('global.navigation.all_galleries'), galleries_path, is_morelink: true)
```
rwd: If you are viewing an individual gallery, the submenu nav item for that gallery is highlighted as current, but so is the nav item for all galleries when it should not be.

app/views/concerns/navigable_view.rb
```diff
@@ -138,6 +140,23 @@ def navigation_global_primary_nav_blog_submenu_items
...
+      text: t('global.navigation.galleries'),
+      is_current: controller.controller_name == 'galleries',
+      submenu: {
+        items: navigation_global_primary_nav_galleries_submenu_items
+      }
+    }
+  end
+
+  def navigation_global_primary_nav_galleries_submenu_items
+    Gallery.published.order(published_at: :desc).map do |gallery|
```
rwd: This will include **all** published galleries in the submenu which will soon become unusable. It should be limited to the _x_ most recent.

PR #849, opened 2017-03-21
Re: #2024 Added collection name to search results model

andyjmaclean:


andyjmaclean:
@rwd #2024 will be testable when this is merged

PR #852, opened 2017-03-22
Filter blogs by theme

rwd:
* Hard-coded fashion option for the equivalent tag on Pro. This and any other options will likely need to remain hard-coded as this app has no way of deducing dynamically which tags from Pro should be in the list of options.
* Also included are a couple of bug fixes for pagination

### DIFF
--- a/app/controllers/blog_posts_controller.rb
+++ b/app/controllers/blog_posts_controller.rb
@@ -15,6 +15,7 @@ class BlogPostsController < ApplicationController
                   where(filters).
                   page(pagination_page).per(pagination_per).all
     @hero_image = homepage_hero_image
+    @theme_filters = theme_filters
     @selected_theme = theme
   end
 
@@ -30,7 +31,7 @@ class BlogPostsController < ApplicationController
 
   def filters
     {}.tap do |filters|
-      filters[:blogs] = theme_filters[theme.to_sym]
+      filters[:blogs] = (theme_filters[theme.to_sym] || {})[:filter]
       filters[:tags] = tag unless tag.nil?
     end
   end
@@ -45,8 +46,14 @@ class BlogPostsController < ApplicationController
 
   def theme_filters
     {
-      all: 'europeana-fashion', # comma-separated list of all blogs to include
-      fashion: 'europeana-fashion'
+      all: {
+        filter: 'europeana-fashion', # comma-separated list of all blogs to include
+        label: t('global.actions.filter-all')
+      },
+      fashion: {
+        filter: 'europeana-fashion',
+        label: Topic.find_by_slug('fashion').label
+      }
     }
   end
 end
--- a/app/controllers/blog_posts_controller.rb
+++ b/app/controllers/blog_posts_controller.rb
@@ -12,15 +12,15 @@ class BlogPostsController < ApplicationController
 
   def index
     @blog_posts = Pro::BlogPost.includes(:network, :persons).
-                    where(blog_post_filters).
-                    page(pagination_page).per(pagination_per).all
+                  where(blog_post_filters).
+                  page(pagination_page).per(pagination_per).all
     @hero_image = homepage_hero_image
     @selected_theme = blog_posts_theme
   end
 
   def show
     @blog_post = Pro::BlogPost.includes(:network, :persons).
-                   where(slug: params[:slug]).first
+                 where(slug: params[:slug]).first
   end
 
   protected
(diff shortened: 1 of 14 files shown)


app/controllers/blog_posts_controller.rb
```diff
@@ -24,4 +25,12 @@ def show
   def pro_blog_posts
     Pro::BlogPost.includes(:network, :persons)
   end
+
+  def themed_blog_posts
```
lbiedinger: There's something peculiar about needing a method like this. Maybe due to the fact that we have to create the pro_blog_posts method in order to create the scope in the first place.

I wonder if it makes sense to move the assignment of the blog_posts for the index action to a before_action like: 
```
  before_action :set_blog_posts, only: :index
```
and then move the logic for determining which posts to retrieve into the set_blog_posts like:
```
  def set_blog_posts
    if blog_posts_theme == 'all'
      @blog_posts = pro_blog_posts.page(pagination_page).per(pagination_per).all
    else
      @blog_posts = pro_blog_posts.where(tags: blog_posts_theme).page(pagination_page).per(pagination_per).all
    end
  end
```
I'm not 100% sure it's stylistically the better option, but it means we wouldn't call themed_blog_posts in a situation where we're not actually retrieving themed blog posts, rather than all blog posts. @rwd What do you think?

app/views/blog_posts/index.rb
```diff
@@ -29,8 +29,25 @@ def blog_items
...
+    def blogs_filter_options
+      {
+        filter_name: 'theme',
+        options: blogs_filter_options_options
+      }
+    end
+
     protected
 
+    def blogs_filter_options_options
```
lbiedinger: can this be renamed to just "filter_options"?

blogs_filter_options_options seems very verbose. If renaming, make sure to change in the method call above.

app/controllers/blog_posts_controller.rb
rwd: @lbiedinger I take your point, but what is not covered by the controller yet is the fact that in the index action we will never be requesting _all_ blog posts from Pro, but will always filter them by some tag (or other Bolt taxonomy term). I'll leave this as-is for the moment and revisit when we have full, categorised, test data to work against.

app/views/blog_posts/index.rb
```diff
@@ -29,10 +29,27 @@ def blog_items
...
+      }
+    end
+
     protected
 
+    def filter_options
+      [
+        { label: t('global.actions.filter-all'), value: 'all' },
+        { label: Topic.find_by_slug('fashion').label, value: 'fashion' }
+      ].tap do |options|
```
lbiedinger: any way we can make use of the theme_filters method from the controller to populate this?

app/views/concerns/paginated_json_api_result_set_view.rb
```diff
@@ -49,8 +49,14 @@ def pagination_total_pages
...
       # Calculate it ourselves instead
-      (pagination_total / pagination_per_page) +
-        ((pagination_total / pagination_per_page).zero? ? 0 : 1)
+      if pagination_total.zero?
+        0
+      elsif pagination_total < pagination_per_page
+        1
+      else
+        (pagination_total / pagination_per_page) +
+          ((pagination_total / pagination_per_page).zero? ? 0 : 1)
```
lbiedinger: is this meant to be % instead of / ?

lbiedinger:
The most obvious note is that the filter by theme option seems pointless while we only have one theme.

Another thing though: Clicking on a tag will remove any applied theme, and likewise applying a theme doesn't remember a selected tag.

Also, the tag display seems messed up, not sure if it's to do with this PR though.

app/views/concerns/paginated_json_api_result_set_view.rb
rwd: Yes! Thanks.

rwd:
@lbiedinger as discussed on Assembla, we will retain the theme filter as it is intended that we will soon have others, but will remove the links from tags for the time being. The presentation of the tags is not a consequence of this PR.

spec/views/blog_posts/index_spec.rb
```diff
@@ -34,4 +41,11 @@
     render
     expect(rendered).to have_selector('.item-preview .category-flag', text: blog_posts.first.taxonomy[:blogs].values.first)
   end
+
+  it 'has theme filter' do
```
lbiedinger: these are failing now

rwd: @lbiedinger thanks. I have spec'd both scenarios now, i.e. theme filters enabled and disabled.

PR #853, opened 2017-03-22
Feat/2036 galleries feed

lbiedinger:


app/views/galleries/index.atom.builder
```diff
@@ -0,0 +1,34 @@
...
+  image_url ||= @hero_image && @hero_image.file.present? ? @hero_image.file.url : nil
+
+  %[
+    <h2>#{gallery.title} <span>(#{image_count} #{t('global.labels.images')})</span></h2>
+    <img src=\"#{image_url}\"/>
+    <p>#{gallery.description}<p>
+  ]
+end
+
+atom_feed do |feed|
```
rwd: To get caching working on the Atom feeds, wrap this `atom_feed` block in:
```ruby
# @todo only insert "views/" into our cache keys when it's needed, i.e. in the context of Mustache view classes
cache(cache_key(@body_cache_key).sub(/\Aviews\//, ''), skip_digest: true) do
  # ...
end
```

We shouldn't really need to do this substitution on the cache key, but we are inserting "views/" into the cache key for the Mustache view classes, whereas the Rails cache helper inserts it itself in the context of view templates like this Atom builder... so we get it repeated. We will want to refactor this when time permits to intelligently insert "views/" only when needed, but for now, the above workaround should suffice here.

app/views/galleries/index.atom.builder
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
```
rwd: If you include the `GalleryDisplayingView` concern here with:
```ruby
self.class.send(:include, GalleryDisplayingView)
```

... then you could reuse its `#presenter_for_gallery_image` and `#document_for_gallery_image`.

app/views/galleries/index.rss.builder
```diff
@@ -0,0 +1,43 @@
...
+      xml.lastBuildDate(@galleries.maximum(:published_at).rfc2822)
+      xml.tag!('atom:link', rel: 'self', type: 'application/rss+xml', href: galleries_url(format: 'rss'))
+
+      @galleries.each do |gallery|
+        xml.item do
+          xml.title(gallery.title)
+          xml.link(gallery_url(gallery))
+          xml.pubDate(gallery.published_at.rfc2822)
+          xml.guid(gallery_url(gallery))
+          xml.author('noreply@europeana.eu (Euroepana Galleries)')
```
rwd: As author is not required in RSS, unlike in Atom, let's just omit this.

app/views/galleries/index.rss.builder
```diff
@@ -0,0 +1,43 @@
...
+    xml.channel do
+      xml.title("Europeana - #{t('global.galleries')}")
+      xml.description(t('site.galleries.description'))
+      xml.link(galleries_url)
+      xml.language(locale.to_s)
+      xml.lastBuildDate(@galleries.maximum(:published_at).rfc2822)
+      xml.tag!('atom:link', rel: 'self', type: 'application/rss+xml', href: galleries_url(format: 'rss'))
+
+      @galleries.each do |gallery|
+        xml.item do
```
rwd: For each topic associated with the gallery, could we add an element to the feed like: `<category>Art</category>`

app/views/galleries/index.rss.builder
```diff
@@ -0,0 +1,43 @@
...
+          xml.title(gallery.title)
+          xml.link(gallery_url(gallery))
+          xml.pubDate(gallery.published_at.rfc2822)
+          xml.guid(gallery_url(gallery))
+          xml.author('noreply@europeana.eu (Euroepana Galleries)')
+          xml.description(entry_summary(gallery))
```
rwd: Let's keep this simple and just put `gallery.description` in here.

app/views/galleries/index.rss.builder
```diff
@@ -0,0 +1,43 @@
...
+  image_count = gallery.images.count
+  unless image_count.zero?
+    presenter = presenter_for_gallery_image(gallery.images.first)
+    image_url = presenter.field_value('edmPreview')
+  end
+  image_url ||= @hero_image && @hero_image.file.present? ? @hero_image.file.url : nil
+
+  %[
+    <h2>#{gallery.title} <span>(#{image_count} #{t('global.labels.images')})</span></h2>
+    <img src=\"#{image_url}\"/>
```
rwd: Could you move this into either an `<enclosure>` or `<media:thumbnail>` on each item?.

app/views/galleries/index.rss.builder
lbiedinger: I've also added the image to the content so that the what's happening feed can parse it out

app/views/galleries/index.rss.builder
lbiedinger: added below

app/views/galleries/index.rss.builder
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+self.class.send(:include, GalleryDisplayingView)
+
+def gallery_image(gallery)
+  unless gallery.images.count.zero?
+    presenter = presenter_for_gallery_image(gallery.images.first)
+    image_url = presenter.field_value('edmPreview')
```
rwd: We should not be using edmPreview as-is, but converting it to the thumbnail API equivalent. Just reuse `GalleryDisplayingView#gallery_image_thumbnail` which does just that.

app/views/galleries/index.rss.builder
```diff
@@ -0,0 +1,43 @@
...
+          xml.link(gallery_url(gallery))
+          xml.pubDate(gallery.published_at.rfc2822)
+          xml.guid(gallery_url(gallery))
+          gallery.categorisations.each do |categorisation|
+            xml.category(categorisation.topic_label)
+          end
+          xml.enclosure(url: gallery_image(gallery), length: 0, type: 'image/*')
+          xml.description(gallery.description)
+          xml.content(:encoded) do |content|
+            content.cdata!("<img src=\"#{gallery_image(gallery)}\"/>")
```
rwd: OK, I understand why you've done this for our internal use case, but I think in future we should adapt our feed consumer to first detect images in dedicated elements like enclosure before looking in content.

app/views/galleries/show.rb
```diff
@@ -11,6 +11,12 @@ def page_title
       mustache[:page_title] ||= [@gallery.title, site_title].join(' - ')
     end
 
+    def head_links
+      mustache[:head_links] ||= begin
+        { items: [{ rel: 'alternate', type: 'application/rss+xml', href: galleries_url(format: 'rss') }] + super[:items] }
```
rwd: This would lose anything else coming from `super` not in the element keyed `:items`

lbiedinger: is there anywhere we're populating anything other than :items?
I've copied this from app/views/portal/show.rb

The only other place the head_links method is being declared as far as I can see is in aplication_view.rb where it always only has :items.

rwd: Yep, fair enough.

PR #860, opened 2017-03-29
Event tweaks

rwd:
* 404 if requested event is not found on Pro
* Add geolocation to event view data, for display as map
* Filters, commented out, pending available data on Pro for testing

### DIFF
--- a/app/controllers/blog_posts_controller.rb
+++ b/app/controllers/blog_posts_controller.rb
@@ -20,11 +20,10 @@ class BlogPostsController < ApplicationController
   end
 
   def show
-    @blog_post = Pro::BlogPost.includes(:network).
-                 where(filters).
-                 where(slug: params[:slug]).first
+    result = Pro::BlogPost.includes(:network).where(filters).where(slug: params[:slug])
+    @blog_post = result.first
 
-    fail JsonApiClient::Errors::NotFound.new(request.original_url) if @blog_post.nil?
+    fail JsonApiClient::Errors::NotFound.new(result.links.links['self']) if @blog_post.nil?
   end
 
   protected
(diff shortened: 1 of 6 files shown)


app/controllers/blog_posts_controller.rb
```diff
@@ -20,11 +20,10 @@ def index
   end
 
   def show
-    @blog_post = Pro::BlogPost.includes(:network).
-                 where(filters).
-                 where(slug: params[:slug]).first
+    result = Pro::BlogPost.includes(:network).where(filters).where(slug: params[:slug])
```
lbiedinger: this should probably be named `results` if it's enumerable

app/controllers/events_controller.rb
```diff
@@ -13,12 +14,29 @@ class EventsController < ApplicationController
   def index
     @events = Pro::Event.includes(:locations, :network).
               order('-end_event').
+              # Uncomment to restrict to events tagged "culturelover"
+              # where(filters).
+              # Uncomment to restrict to current and future events
+              # where(end_event: ">=" + Date.today.strftime).
```
lbiedinger: this will need to be conditional once we use it, to take into account looking up past events
`where(end_event: "<" + Date.today.strftime).` < if that's the reverse query it's probably enough to add a ternary around the less/greater-or-equal operator.

app/controllers/events_controller.rb
```diff
@@ -13,12 +14,29 @@ class EventsController < ApplicationController
...
               page(pagination_page).per(pagination_per).all
     @hero_image = homepage_hero_image
   end
 
   def show
-    @event = Pro::Event.includes(:locations, :network).
-             where(slug: params[:slug]).first
+    result = Pro::Event.includes(:locations, :network).
```
lbiedinger: again should probably be `results`

PR #868, opened 2017-03-31
Filtering blog posts and events

rwd:
* Events and blog posts from Pro JSON API are now always filtered by tag: "culturelover" for all, "culturelover-fashion" for Fashion
* Labels on tiles on listing pages are taken from any "culturelover-theme" tags, with a lookup for a label in the `Topic` model, where "theme" from the tag needs to match a topic slug
* Events can be filtered by theme, but this is not exposed unless `ENABLE_EVENTS_THEME_FILTER` is set
* Events default to future events only, but can be switched to show past, although the front-end does not respect the field name defined for this

### DIFF
--- a/app/presenters/pro_resource_presenter.rb
+++ b/app/presenters/pro_resource_presenter.rb
@@ -39,12 +39,12 @@ class ProResourcePresenter
 
   # We omit from view internal-use tags starting "culturelover"
   def displayable_tags
-    return nil unless resource.has_taxonomy?(:tags) 
+    return nil unless resource.has_taxonomy?(:tags)
     @displayable_tags ||= resource.taxonomy[:tags].reject { |_pro_path, tag| tag.start_with?('culturelover') }
   end
 
   def theme_label_tags
-    return nil unless resource.has_taxonomy?(:tags) 
+    return nil unless resource.has_taxonomy?(:tags)
     @theme_label_tags ||= resource.taxonomy[:tags].select { |_pro_path, tag| tag.start_with?('culturelover-') }
   end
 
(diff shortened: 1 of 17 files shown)


app/presenters/pro_resource_presenter.rb
```diff
@@ -37,9 +37,15 @@ def has_displayable_tags?
     displayable_tags.present?
   end
 
-  # We omit from view the "culturelover" internal-use tag
+  # We omit from view internal-use tags starting "culturelover"
   def displayable_tags
-    resource.taxonomy[:tags].reject { |_pro_path, tag| tag == 'culturelover' }
+    return nil unless resource.has_taxonomy?(:tags)
+    @displayable_tags ||= resource.taxonomy[:tags].reject { |_pro_path, tag| tag.start_with?('culturelover') }
```
lbiedinger: might make sense for this to make use of pro_json_api_filters[:tags] to allow more flexibility 
not sure if that's always guaranteed accessible at the moment since that's defined in the pro_json_api_consumer concern

app/views/concerns/theme_filterable_view.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+##
+# Filtering by theme
+module ThemeFilterableView
+  extend ActiveSupport::Concern
+
+  protected
+
+  def theme_filter_options
+    theme_options = theme_filters.map { |key, data| { label: data[:label], value: key } }.tap do |options|
```
lbiedinger: is the reason not to use 'pro_json_api_theme_filters' here, to keep it more generic so we can reuse this in for example galleries or somewehre?

rwd: Yep, exactly that.

PR #882, opened 2017-04-19
Re: #2106 Gallery tooltips

andyjmaclean:


andyjmaclean:
Locally none of my published galleries show, and the presence of {{label}} in the template for a gallery foyer item is causing the back end to crash.  These issues are unrelated to this commit, however.

andyjmaclean:
@rwd the latest Gemref should be used

app/views/galleries/index.rb
```diff
@@ -91,7 +98,19 @@ def published_gallery_topic_options
     end
 
     def galleries_content
-      @galleries.map { |gallery| gallery_content(gallery) }
+      res = @galleries.map { |gallery| gallery_content(gallery) }
+      if res.length > 2
```
rwd: Galleries do not necessarily have a description, so this "clicktip" may be pointing to... nothing, i.e. no _i_ symbol on the second gallery if it has no description. You need to check if the second has a description.

app/views/galleries/index.rb
```diff
@@ -91,7 +98,19 @@ def published_gallery_topic_options
     end
 
     def galleries_content
-      @galleries.map { |gallery| gallery_content(gallery) }
+      res = @galleries.map { |gallery| gallery_content(gallery) }
```
rwd: Could you restore `galleries_content` to what it was before, then move this clicktip-adorning code into a separate method, that you might call like `add_info_clicktip(galleries_content))`?

rwd:
@andyjmaclean you should merge develop branch into this feature branch and push that to resolve the Gemfile conflicts.

andyjmaclean:
@rwd I'm using 'break' to exit a loop and the hound is warning 'use next to skip iteration'.   This is functionally different to what's needed - I want to break (not just skip).

Can we ignore this warning?

rwd:
@andyjmaclean that warning is still there. Hound is suggesting you do `next if content[location][:info].present?` instead of enclosing the following lines in an `if` block.

However, I'd be inclined to rewrite your `add_info_clicktip` as:
```ruby
    def info_clicktip_target(galleries_content)
      if galleries_content[1].present? && galleries_content[1][:info].present?
        galleries_content[1]
      else
        galleries_content.detect { |gc| gc[:info].present? }
      end
    end

    def add_info_clicktip(galleries_content)
      info_clicktip_target(galleries_content).tap do |target|
        unless target.nil?
          target[:info_clicktip] = {
            clicktip: {
              activator:    '.image-set:eq(1) .svg-icon-info',
              direction:    'top',
              id:           'gallery-hover-info',
              persistent:   true,
              tooltip_text: t('global.tooltips.channels.galleries.info')
            }
          }
        end
      end

      galleries_content
    end
```

... which I think is tidier and more readable.

PR #888, opened 2017-05-04
re #2086 detect pro events and blogs

lbiedinger:


app/models/concerns/has_external_urls.rb
```diff
@@ -41,4 +41,12 @@ def twitter?
   def europeana_blog?
     url_in_domain?('blog.europeana.eu')
   end
+
+  def pro_blog?
+    !(url =~ %r(://([^/]*.)?europeana.eu/portal/(.*)blogs.rss(.*))).nil?
```
rwd: 1. `(://([^/]*.)?`: this being optional (via `?`) at the start of the RegExp makes it achieve nothing
2. `/portal/(.*)blogs`: if the `(.*)` is here only to catch different language codes, replace it with `[a-z]{2}/`
3. `blogs.rss`: escape the `.`
4. `rss(.*)`: this at the end of this RegExp achieves nothing

app/models/concerns/has_external_urls.rb
```diff
@@ -41,4 +41,12 @@ def twitter?
...
+  end
+
+  def pro_events?
+    !(url =~ %r(://([^/]*.)?europeana.eu/portal/(.*)events.rss(.*))).nil?
```
rwd: See comment about `#pro_blog?`

PR #890, opened 2017-05-08
use FacetLinkGroupPresenter for preview search label

lbiedinger:
The changes made are using FacetLinkGroupPresenter instead of calling the FacetPresenter directly since the FacetPresenter needs a Europeana::Blacklight::Response::Facets::FacetItem and the Europeana::Blacklight::Response::Facets::FacetField the FacetLinkGroupPresenter works with the FacetEntry directly so it keeps the facet_entry_display_value method cleaner.

### DIFF
--- a/app/views/collections/show.rb
+++ b/app/views/collections/show.rb
@@ -126,10 +126,10 @@ module Collections
 
     def carousel_data
       case @landing_page.settings[:layout_type]
-        when 'default'
-          helpers.tumblr_feed_content(@landing_page)
-        when 'browse'
-          helpers.page_feeds_content(@landing_page)
+      when 'default'
+        helpers.tumblr_feed_content(@landing_page)
+      when 'browse'
+        helpers.page_feeds_content(@landing_page)
       end
     end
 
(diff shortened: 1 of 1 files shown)


rwd:
@lbiedinger given that for each facet entry, a `FacetPresenter` instance is already created, I think we ought to reuse that rather than also create a `FacetLinkGroupPresenter` instance. Something like this:
```ruby
    def preview_search_data
      return nil if @landing_page.facet_entries.blank?

      @landing_page.facet_entries.map do |facet_entry|
        presenter = facet_entry_presenter(facet_entry)
        {
          preview_search_title: presenter.facet_item_label(entry.facet_value) || facet_entry.title,
          preview_search_type: presenter.facet_title || facet_entry.facet_field,
          preview_search_url: browse_entry_url(facet_entry, @landing_page, format: 'json'),
          preview_search_more_link: browse_entry_url(facet_entry, @landing_page)
        }
      end
    end

    def facet_entry_presenter(facet_entry)
      ff = Europeana::Blacklight::Response::Facets::FacetField.new(facet_entry.facet_field, [])
      FacetPresenter.build(ff, controller)
    end
```

app/views/collections/show.rb
```diff
@@ -172,25 +172,26 @@ def stylised_collection_stats
...
         }
       end
     end
 
-    def facet_entry_field_title(facet_entry)
-      ff = Europeana::Blacklight::Response::Facets::FacetField.new(facet_entry.facet_field, [])
-      presenter = FacetPresenter.build(ff, controller)
-      presenter.facet_title || facet_entry.facet_field
+    def facet_entry_presenter(facet_entry)
+      @facet_presenters ||= {}
```
rwd: If this method is only ever called by `preview_search_data` where the presenter is assigned to a variable, then I'm not sure this object caching gains us anything?

lbiedinger: Because we send the complete list of possible preview searches to the frontend, with the optimisation, It'll only generate a presenter for each facet field. That's three presenters that are initialised. Without it the optimisation, it will initialise a presenter for each facet value, so 36 presenters get initialised. (provided 3 facet fields are configured and they all contain 6 values, which would be the default)

rwd: Right you are!

PR #891, opened 2017-05-09
Feat/1804 iframe test

lbiedinger:


config/initializers/europeana.rb
```diff
@@ -17,3 +17,5 @@
   ENV['EUROPEANA_MEDIA_PROXY'] = ENV['EDM_IS_SHOWN_BY_PROXY']
 end
 Rails.application.config.x.europeana_media_proxy = ENV['EUROPEANA_MEDIA_PROXY'] || 'http://proxy.europeana.eu'
+
+Rails.application.config.x.europeana_1914_1918_url = ENV['EUROPEANA_1914_1918_URL'] || 'http://acceptance.1418.eanadev.org'
```
rwd: This should default to the new hostname assigned to the production deployment of 1914-1918, but that has not yet been decided. (See https://europeanadev.assembla.com/spaces/europeana-npc/tickets/2151.)

app/views/collections/ugc.rb
```diff
@@ -0,0 +1,26 @@
...
+          name: 'pageName', value: 'e7a_1418'
+        }
+      ]
+    end
+
+    def page_title
+      'Europeana - First World War'
+    end
+
+    def content
```
rwd: Shouldn't these URLs go into `#navigation` rather than `#content`?

app/views/collections/ugc.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    def js_vars
+      [
+        {
+          name: 'pageName', value: 'e7a_1418'
+        }
+      ]
+    end
+
+    def page_title
```
rwd: Could you instead implement `#page_content_heading` returning just the title of the firstworldwar collection if it exists, otherwise falling back to hard-coded  'First World War', leaving the addition of the site title to `ApplicationView#page_title`.

.env.example
```diff
@@ -133,3 +133,6 @@
 
 ## Relative URL root at which the app is deployed.
 # RAILS_RELATIVE_URL_ROOT=/portal
+
+## The URL to use for pulling in embeded 1914-1918 UGC funtionality.
```
rwd: "embedded"

config/initializers/europeana.rb
lbiedinger: yeah, same goes for the example in .env.example

app/views/collections/ugc.rb
lbiedinger: Doesn't really matter as far as I'm concerned. The base_1418_url does however determine where the content of the iframe gets loaded in from. So from that point of view it pertains to the content.
I've removed the portal_url as there was a misunderstanding between myself and Andy as to what it's purpose was and it wasn't doing what Andy intended it to do in the first place.

rwd: Fair enough. I thought these were used for breadcrumbs, but if not, then remaining in `#content` makes sense.

app/views/collections/ugc.rb
```diff
@@ -0,0 +1,40 @@
...
+      [
+        {
+          name: 'pageName', value: 'e7a_1418'
+        }
+      ]
+    end
+
+    def page_content_heading
+      if Collection.find_by_key('firstworldwar').present?
+        Collection.find_by_key('firstworldwar').title
```
rwd: Let's not make the same db query three times...

Looking at `CollectionsController#ugc`, the collection should now be in `@collection`.

app/views/collections/ugc.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    def content
+      {
+        base_1418_url: config.x.europeana_1914_1918_url,
+      }
+    end
+
+    def collection_data
+      mustache[:collection_data] ||= begin
+        collection = @collection
```
rwd: What's the point of this? Just use `@collection` in the following lines...

lbiedinger: no point, other than this was copy-pasted. Will update.

PR #896, opened 2017-05-16
Update mail and nokogiri libraries

rwd:


rwd:
Included in https://github.com/europeana/europeana-portal-collections/pull/897

PR #900, opened 2017-05-23
Feat/2128 federated search

lbiedinger:
https://europeanadev.assembla.com/spaces/europeana-npc/tickets/2128

### DIFF
--- a/app/controllers/federations_controller.rb
+++ b/app/controllers/federations_controller.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+##
+# Provides Blacklight search and browse, within a content Collection
+class FederationsController < ApplicationController
+  # The federated action which is used to retrieve federated results via the foederati gem.
+  def show
+    @collection = find_collection
+    provider = params[:id]
+    @query = params[:query]
+    if @collection.settings_federated_providers && @collection.settings_federated_providers.detect(provider.to_sym)
+      foederati_provider = Foederati::Providers.get(provider.to_sym)
+      @federated_results = Foederati.search(provider.to_sym, query: @query)[provider.to_sym]
+      @federated_results[:more_results_label] = "View more at #{foederati_provider.display_name}"
+      @federated_results[:more_results_url] = format(foederati_provider.urls.site, query: @query)
+      @federated_results[:tab_subtitle] = "#{@federated_results[:total]} Results"
+
+      @federated_results[:search_results] = @federated_results.delete(:results)
+      @federated_results[:search_results].each do |result|
+        result[:img] = { src: result.delete(:thumbnail) } if result[:thumbnail]
+        result[:object_url] = result.delete(:url)
+      end
+    end
+    render json: @federated_results
+  end
+
+  def find_collection
+    Collection.find_by_key!(params[:collection]).tap do |collection|
+      authorize! :show, collection
+    end
+  end
+end
--- a/app/controllers/federations_controller.rb
+++ b/app/controllers/federations_controller.rb
@@ -6,8 +6,9 @@ class FederationsController < ApplicationController
   def show
     @collection = find_collection
     provider = params[:id]
-    @query = params[:query]
-    if @collection.settings_federated_providers && @collection.settings_federated_providers.detect(provider.to_sym)
+    federation_config =  @collection.federation_configs.where(provider: provider).first
+    @query = "#{params[:query]} #{federation_config.context_query}"
+    if @collection.federation_configs && federation_config
       foederati_provider = Foederati::Providers.get(provider.to_sym)
       @federated_results = Foederati.search(provider.to_sym, query: @query)[provider.to_sym]
       @federated_results[:more_results_label] = t('global.actions.view-more-at') + foederati_provider.display_name
--- a/app/controllers/federations_controller.rb
+++ b/app/controllers/federations_controller.rb
@@ -6,7 +6,7 @@ class FederationsController < ApplicationController
   def show
     @collection = find_collection
     provider = params[:id]
-    federation_config =  @collection.federation_configs.where(provider: provider).first
+    federation_config = @collection.federation_configs.where(provider: provider).first
     @query = "#{params[:query]} #{federation_config.context_query}"
     if @collection.federation_configs && federation_config
       foederati_provider = Foederati::Providers.get(provider.to_sym)
--- a/app/controllers/federations_controller.rb
+++ b/app/controllers/federations_controller.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 ##
 # Provides Federated content via JSON, to be displayed within a content Collection
 class FederationsController < ApplicationController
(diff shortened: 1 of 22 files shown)


app/controllers/federations_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+class FederationsController < ApplicationController
+  # The federated action which is used to retrieve federated results via the foederati gem.
+  def show
+    @collection = find_collection
+    provider = params[:id]
+    federation_config = @collection.federation_configs.where(provider: provider).first
+    @query = "#{params[:query]} #{federation_config.context_query}"
+    if @collection.federation_configs && federation_config
+      foederati_provider = Foederati::Providers.get(provider.to_sym)
+      @federated_results = Foederati.search(provider.to_sym, query: @query)[provider.to_sym]
```
rwd: Lines 14-23 here should probably go into a JSON builder view.

app/views/portal/index.rb
```diff
@@ -185,6 +185,26 @@ def facets_selected
       end
     end
 
+    def federated_search_enabled
+      @collection && @collection.federation_configs.count.positive?
```
rwd: Wouldn't `federation_configs.present?` suffice?

app/views/portal/index.rb
```diff
@@ -185,6 +185,26 @@ def facets_selected
...
+    end
+
+    def federated_search_conf
+      mustache[:federated_search_conf] ||= begin
+        {
+          tab_items: @collection.federation_configs.map do |config|
```
rwd: Can you move the content of this block into its own method, for readability?

app/views/portal/index.rb
```diff
@@ -185,6 +185,26 @@ def facets_selected
...
+            foederati_provider = Foederati::Providers.get(config.provider)
+            next unless foederati_provider
```
rwd: Instead of this, you could use:
```ruby
@collection.federation_configs.select { |config| Foederati::Providers.get(config.provider).present? }.map do |config|
```
... so the `next` line can be removed, as can the `#reject` call at the block's end.

config/initializers/foederati.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# config/initializers/foederati.rb
```
rwd: This can be removed.

app/models/federation_config.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  delegate :provider_enum, to: :class
+
+  class << self
+    def provider_enum
+      Foederati::Providers.registry.keys
+    end
+  end
+
+  validates_presence_of %i{collection_id provider}
```
rwd: To keep this consistent, can we use the `validates` format for the presence validator too?

app/models/federation_config.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class FederationConfig < ActiveRecord::Base
+  belongs_to :collection
+
+  delegate :provider_enum, to: :class
+
+  class << self
+    def provider_enum
+      Foederati::Providers.registry.keys
```
rwd: minus Europeana

app/controllers/federations_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+##
+# Provides Federated content via JSON, to be displayed within a content Collection
+class FederationsController < ApplicationController
+  # The federated action which is used to retrieve federated results via the foederati gem.
+  def show
+    @collection = find_collection
+    provider = params[:id]
+    federation_config = @collection.federation_configs.where(provider: provider).first
+    @query = "#{params[:query]} #{federation_config.context_query}"
```
rwd: Tweaking this to: `[params[:query], federation_config.context_query].compact.join(' ')` would prevent needless spaces going into the query sent to the provider when the user enters an empty search, e.g. http://localhost:3000/portal/en/collections/music?q=&view=grid

db/migrate/20170531140414_create_federation_config.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class CreateFederationConfig < ActiveRecord::Migration
+  def self.up
+    create_table :federation_configs do |t|
+      t.integer :collection_id, null: false # The ID of the collection this config applies to
+      t.string :provider, null: false       # The name of the provider this config applies to
+      t.string :context_query               # The default query which should always be applied.
```
rwd: I think just "query" would have sufficed for this column name. Collection + provider = context.

app/controllers/federations_controller.rb
lbiedinger: empty searches are a bit problematic (specifically for trove/dpla) when combined with a "context" I'm trying to figure out a better solution for this.

app/models/federation_config.rb
lbiedinger: do you mean:
```
  validates :collection, presence: true
  validates :provider, presence: true
```
?

db/migrate/20170531140414_create_federation_config.rb
lbiedinger: generally yes, however I'd disagree on this since "query" is already used by both the portal and the foederati gem to refer to the search term/the final search string. I think it's nicer to differentiate this from the rest of the "query" as it is only a partial string of the final "query".

app/models/federation_config.rb
rwd: Yes, which can of course be compacted into `validates :collection_id, :provider, presence: true`

app/views/federations/show.json.jbuilder
```diff
@@ -0,0 +1,12 @@
+
```
rwd: This is Ruby, so should get the frozen_string_literal magic comment we have in .rb files.

app/views/federations/show.json.jbuilder
```diff
@@ -0,0 +1,12 @@
...
+@federated_results[:more_results_label] = t('global.actions.view-more-at') + @foederati_provider.display_name
```
rwd: This doesn't seem quite the right approach to a jbuilder view. I'd expect it not to modify `@federated_results` and to make fuller use of the jbuilder DSL, like:
```ruby
json.total @federated_results[:total]
json.more_results_label t('global.actions.view-more-at') + @foederati_provider.display_name
json.more_results_url format(@foederati_provider.urls.site, query: @query)
json.tab_subtitle [number_with_delimiter(@federated_results[:total]), t('site.results.results')].join(' ')
json.search_results @federated_results[:results] do |result|
  json.title result[:title]
  json.img { src: result[:thumbnail] } if result[:thumbnail]
  json.object_url result[:url]
end
```

Also, the present implementation leaves blank `:thumbnail` entries on `:search_results`, which the above would not.

lbiedinger: agreed, this is much nicer than the copy paste job I did.
`json.img { src: result[:thumbnail] }` 
needs to be 
` json.img { json.src result[:thumbnail] }`

PR #906, opened 2017-06-01
Re: #2160 Add UGC navbar to object page

andyjmaclean:


app/views/portal/show.rb
```diff
@@ -83,7 +83,11 @@ def content
           refs_rels: presenter.field_group(:refs_rels),
           similar: similar_items,
           named_entities: named_entities,
-          thumbnail: field_value('europeanaAggregation.edmPreview', tag: false)
+          thumbnail: field_value('europeanaAggregation.edmPreview', tag: false),
+          ugc_content: {
+            base_1418_url: 'http://ec2-54-154-38-184.eu-west-1.compute.amazonaws.com',
```
lbiedinger: instead of hardcoding here, use `config.x.europeana_1914_1918_url`

andyjmaclean:
@rwd the recommended view relies on @collection which is always null in this context

andyjmaclean:
Added a parameter to ugc_displaying view allowing it to be forced when no collection is present

app/views/concerns/ugc_content_displaying_view.rb
```diff
@@ -7,8 +7,8 @@ module UgcContentDisplayingView
 
   protected
 
-  def ugc_content
-    if @collection && @collection.accepts_ugc?
+  def ugc_content(force)
```
lbiedinger: `def ugc_content(force = false)`

PR #907, opened 2017-06-06
order federated tabs, rubocop

lbiedinger:


app/views/portal/index.rb
```diff
@@ -198,14 +200,15 @@ def federated_search_conf
     end
 
     def federated_tab_items
-      @collection.federation_configs.select { |config| Foederati::Providers.get(config.provider).present? }.map do |config|
+      formated_items = @collection.federation_configs.select { |config| Foederati::Providers.get(config.provider).present? }.map do |config|
```
rwd: Could this be renamed `formatted_items`?

PR #923, opened 2017-06-20
use zero promo tile for og:image on browse landing pages

lbiedinger:
For browse landing pages use the image from the promo tile at position 0, if this isn't found for any reason or the landing page is using the default display, default to the hero image.
+specs

### DIFF
--- a/app/views/collections/show.rb
+++ b/app/views/collections/show.rb
@@ -25,7 +25,7 @@ module Collections
           { meta_property: 'og:url', content: collection_url(@collection.key) }
         ]
         head_meta << { meta_property: 'og:title', content: title } unless title.nil?
-        head_meta << { meta_property: 'og:image', content: og_image } unless og_image.blank?
+        head_meta << { meta_property: 'og:image', content: @landing_page.og_image } unless @landing_page.og_image.blank?
         head_meta + super
       end
     end
@@ -102,21 +102,6 @@ module Collections
     end
 
     private
-
-    def og_image
-      og_image_from_promo || og_image_from_hero
-    end
-
-    def og_image_from_promo
-      return unless @landing_page.settings_layout_type == 'browse'
-      promo = @landing_page.promotions.find_by(position: 0)
-      promo && promo.file.present? ? promo.file.url : nil
-    end
-
-    def og_image_from_hero
-      hero_conf = hero_config(@landing_page.hero_image)
-      hero_conf[:hero_image] unless hero_conf.nil? || hero_conf[:hero_image].blank?
-    end
 
     def head_meta_description
       mustache[:head_meta_description] ||= begin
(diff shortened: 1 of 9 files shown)


app/views/collections/show.rb
```diff
@@ -102,6 +103,21 @@ def version
...
+
+    def og_image_from_promo
+      return unless @landing_page.settings_layout_type == 'browse'
+      promo = @landing_page.promotions.find_by(position: 0)
+      promo && promo.file.present? ? promo.file.url : nil
+    end
+
+    def og_image_from_hero
+      hero_conf = hero_config(@landing_page.hero_image)
+      URI.join(root_url, hero_conf[:hero_image]) unless hero_conf.nil? || hero_conf[:hero_image].blank?
```
rwd: Why are we doing `URI.join(root_url, hero_conf[:hero_image])` when the images are stored on S3 in our CF deployments?

app/views/collections/show.rb
```diff
@@ -102,6 +103,21 @@ def version
 
     private
 
+    def og_image
```
rwd: Let's move this onto the `Page::Landing` model.

app/views/collections/show.rb
lbiedinger: I can't see any other reason than that this logic was copy pasted through various views.
I'll remove in all places where this is happening.

app/views/collections/show.rb
lbiedinger: do you mean only the og_image method? if I move og_image_from_promo and og_image_from_hero as well I won't have access to the view concern for hero_configs, unless I refactor that into a general concern, but that'll mean refactoring the portal index view too. Do you think it's worth doing so?

app/views/collections/show.rb
lbiedinger: actually the original reason was probably implemented as a fallback for when the hero image file was empty, so we were returning the root_url, which kind of doesn't make sense since that won't be an image anyway.
Now with the other checks and the hero_config code we have I think we'll not need to do this fallback.

app/views/collections/show.rb
rwd: No, all three methods. All you need from `hero_config` is `:hero_image` which is `hero_image.file.present? ? hero_image.file.url : nil`, and therefore available to the `Page::Landing` model.

app/views/collections/show.rb
```diff
@@ -24,7 +25,7 @@ def head_meta
           { meta_property: 'og:url', content: collection_url(@collection.key) }
         ]
         head_meta << { meta_property: 'og:title', content: title } unless title.nil?
-        head_meta << { meta_property: 'og:image', content: URI.join(root_url, hero[:hero_image]) } unless hero.nil?
+        head_meta << { meta_property: 'og:image', content: @landing_page.og_image } unless @landing_page.og_image.blank?
```
rwd: This potentially results in all of the logic in `Page::Landing#og_image_from_promo` and `Page::Landing#og_image_from_hero` being run twice.

Could we avoid that by either storing `@og_image` on `Page::Landing` after `#og_image` is first called, or storing in a local variable in `Collections::Show#head_meta`?

PR #924, opened 2017-06-21
Use document count for Search API response pagination

lbiedinger:
Re https://europeanadev.assembla.com/spaces/europeana-npc/tickets/1729

### DIFF
--- a/app/views/concerns/paginated_json_api_result_set_view.rb
+++ b/app/views/concerns/paginated_json_api_result_set_view.rb
@@ -59,4 +59,8 @@ module PaginatedJsonApiResultSetView
       end
     end
   end
+
+  def pagination_page_item_count
+    paginated_set.count
+  end
 end
--- a/app/views/concerns/paginated_json_api_result_set_view.rb
+++ b/app/views/concerns/paginated_json_api_result_set_view.rb
@@ -59,8 +59,4 @@ module PaginatedJsonApiResultSetView
       end
     end
   end
-
-  def pagination_page_item_count
-    paginated_set.count
-  end
 end
--- a/app/views/concerns/paginated_view.rb
+++ b/app/views/concerns/paginated_view.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 ##
-# Including view are expected to implement `pagination_set`, returning the set
+# Including view are expected to implement `paginated_set`, returning the set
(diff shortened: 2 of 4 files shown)


app/views/concerns/paginated_json_api_result_set_view.rb
```diff
@@ -59,4 +59,8 @@ def pagination_total_pages
       end
     end
   end
+
+  def pagination_page_item_count
```
rwd: This is not needed (see comment on `PaginatedView#pagination_page_item_count`)

app/views/concerns/paginated_view.rb
```diff
@@ -66,7 +66,8 @@ def pagination_total_pages
   end
 
   def pagination_page_item_count
-    paginated_set.count
+    return paginated_set['itemsCount'] unless paginated_set['itemsCount'].nil?
```
rwd: The `PaginatedView` concern should not need to know about the API response's use of `itemsCount`. If anything this would have belonged in a method override in the `Portal::Index` view.

However, all that's needed to fix pagination for all including views is to rewrite this method:
```ruby
def pagination_page_item_count
  paginated_set.docs.count
end
```

app/views/concerns/paginated_json_api_result_set_view.rb
lbiedinger: If I don't override this and implement either what is currently in the concern or your suggestion, the json responses will throw an exception:
`undefined method 'docs' for #<JsonApiClient::ResultSet:0x005626b49f7b40>`
Either way we'll need to override the method somewhere between the search results and the JSON API repsonses, they're structured differently.
Personally I think the default pagniated_view concern method should work for search results, as that is the portal's primary functionality, then we can just leave this override here.
The other place the pagination is used is in the galleries, which actually don't ever check for the paginated result number, however if they did, the same override as here `paginated_set.count` would work for them, but `paginated_set.docs.count` wouldn't. That's an argument for using `paginated_set.count` in the pagniated_view concern, and overriding the portal/index.rb view's method instead.
What do you think?

rwd: Right... yes, let's do that, i.e. keep `paginated_set.count` in `PaginatedView#pagination_page_item_count` and override in `Portal::Index` with `paginated_set.docs.count`.

PR #932, opened 2017-07-07
Feat/2291

kgish:
Initial changes, stable and tests passing.

### DIFF
--- a/app/controllers/entities_controller.rb
+++ b/app/controllers/entities_controller.rb
@@ -1,14 +1,34 @@
 # frozen_string_literal: true
 class EntitiesController < ApplicationController
+  include Europeana::EntitiesApiConsumer
+
   def suggest
-    render json: Europeana::API.entity.suggest(entity_api_params(text: params[:text]))
+    render json: Europeana::API.entity.suggest(entities_api_suggest_params(params[:text]))
   end
 
-  protected
+  def show
+    authorize! :show, :entity
+    @entity = Europeana::API.
+              entity.fetch(entities_api_fetch_params(params[:type], params[:namespace], params[:identifier]))
 
-  def entity_api_params(query = {})
-    query.merge(scope: 'europeana').tap do |api_params|
-      api_params[:wskey] = ENV['EUROPEANA_ENTITIES_API_KEY'] if ENV['EUROPEANA_ENTITIES_API_KEY']
+    @items_by_query = build_query_items_by(params)
+
+    respond_to do |format|
+      format.html
+      format.json { render json: @entity }
     end
+  end
+
+  private
+
+  def build_query_items_by(params)
+    suffix = "#{params[:type]}/#{params[:namespace]}/#{params[:identifier]}"
+    creator = build_proxy_dc('creator', 'http://data.europeana.eu', suffix)
+    contributor = build_proxy_dc('contributor', 'http://data.europeana.eu', suffix)
+    "#{creator} OR #{contributor}"
+  end
+
+  def build_proxy_dc(name, url, suffix)
+    "proxy_dc_#{name}:\"#{url}/#{suffix}\""
   end
 end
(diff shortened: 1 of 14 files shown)


app/controllers/portal_controller.rb
```diff
@@ -35,7 +35,8 @@ def index
         render json: {
           search_results: @document_list.map do |doc|
             Document::SearchResultPresenter.new(doc, self, @response).content
-          end
+          end,
+          total: @response.total
```
lbiedinger: let's add a test for this to spec/controllers/portal_controller_spec.rb

app/helpers/entities_helper.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+##
+# Entities helpers
+#
+module EntitiesHelper
+  # Capitalize all words in sentence
+  def capitalize_words(sentence)
```
lbiedinger: shouldn't we be able to leave this to the css?

kgish: Okay.

app/controllers/concerns/europeana/entities.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Europeana
+  module Entities
+    extend ActiveSupport::Concern
+    include EntitiesHelper
+
+    protected
+
+    def build_query_items_by(params)
```
lbiedinger: Do we need an Entities concern? are you planing on using build_query_items_by somewhere other than the entities_controller?

app/controllers/entities_controller.rb
```diff
@@ -1,14 +1,27 @@
...
   def suggest
-    render json: Europeana::API.entity.suggest(entity_api_params(text: params[:text]))
+    render json: Europeana::API.entity.suggest(entities_api_suggest_params(params[:text]))
   end
 
-  protected
+  def show
+    @entity = Europeana::API.
+              entity.fetch(entities_api_fetch_params(params[:type], params[:namespace], params[:identifier])).
+              merge(__params__: { type: params[:type], namespace: params[:namespace], identifier: params[:identifier] })
```
lbiedinger: The only place these __params__ are being used is to generate the items_about path, which isn't ready yet and once it will be will probably be more akin to how the items_by is generated. So let's get rid of this and remove the references from the content method in entities/show.rb

app/views/concerns/entity_displaying_view.rb
```diff
@@ -0,0 +1,261 @@
...
+        'Remaining text goes here and is a little bit longer'
+      ],
+      href: 'javscript:alert("Go to the external page")'
+    }
+  end
+
+  private
+
+  # Returns a string
+  def get_entity_value_by_locale(list)
```
lbiedinger: no need to preface this and the following methods with `get_`

kgish: Done.

app/helpers/entities_helper.rb
kgish: Done.

app/views/entities/show.rb
```diff
@@ -0,0 +1,88 @@
...
+    #       { meta_property: 'og:url', content: collection_url(@collection.key) }
+    #     ]
+    #     head_meta << { meta_property: 'og:title', content: title } unless title.nil?
+    #     head_meta << { meta_property: 'og:image', content: @landing_page.og_image } unless @landing_page.og_image.blank?
+    #     head_meta + super
+    #   end
+    # end
+
+    def content
+      params = entity_params
```
kgish: Done.

app/controllers/concerns/europeana/entities.rb
kgish: Done.

app/controllers/entities_controller.rb
kgish: Done.

app/controllers/portal_controller.rb
kgish: Done.

PR #934, opened 2017-07-20
Feat/1877 annotations ajax

lbiedinger:


Gemfile.lock
```diff
@@ -330,8 +330,8 @@ GEM
       thor (~> 0.19.1)
     formatador (0.2.5)
     gli (2.15.0)
-    globalid (0.4.0)
```
rwd: Are these changes to Gemfile.lock intentional? If so, what are they for? If not, could they be reverted?

app/controllers/portal_controller.rb
```diff
@@ -85,6 +96,15 @@ def media
     end
   end
 
+  # GET /record/:id/annotations
+  def annotations
+    _response, document = fetch(doc_id)
```
rwd: This results in a superfluous call to the Record API and should not be necessary solely to retrieve annotations.

app/views/portal/annotations.json.erb
```diff
@@ -0,0 +1 @@
+<%= raw record_annotations.to_json %>
```
rwd: This seems unnecessary. Couldn't this just be rendered from the controller?

app/helpers/record_helper.rb
```diff
@@ -59,6 +59,19 @@ def record_media_items(record, options = {})
     Document::RecordPresenter.new(record, controller).media_web_resources(options).map(&:media_item)
   end
 
+  def record_annotations
```
rwd: Could we add an argument to this method to accept the annotations object instead of presuming it will be in `@annotations`, to make it more reusable?

app/views/portal/annotations.json.erb
kgish:
```
respond_to do |format|
  format.html
  format.json { render json: @annotations }
end
```

rwd: It would still need to run the annotations through the `record_annotations` method, so on second thoughts let's just leave this simple view as-is so as not to be calling view-related methods in the controller.

Gemfile.lock
lbiedinger: not intentional, they're a result of an oversight when merging. It shouldn't make a difference but let's go with the later version for these couple of gems.

app/controllers/concerns/europeana/annotations_api_consumer.rb
```diff
@@ -20,16 +20,10 @@ def document_annotations(document_id)
       nil
     end
 
-    # @todo remove target_id fallback when target_uri field name change deployed
-    #   to production Annotations API
     def annotations_search_for_document(document_id)
       Europeana::API.annotation.search(annotations_api_search_params(document_id))
     rescue Europeana::API::Errors::ServerError => error
```
rwd: This `rescue` and the following `raise` do nothing now and can be removed.

PR #937, opened 2017-07-24
Convert entity_label return value to an array

kgish:
entity_label returns a string which needs to be converted into an array in order for any? to work correctly.

### DIFF
--- a/app/presenters/document/field_group_presenter.rb
+++ b/app/presenters/document/field_group_presenter.rb
@@ -133,7 +133,7 @@ module Document
           if section[:entity_extra].present?
             possible_entities = entities(section[:entity_name], section[:entity_proxy_field])
             salient_entity = possible_entities.detect do |entity|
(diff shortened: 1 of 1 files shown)


app/presenters/document/field_group_presenter.rb
```diff
@@ -133,7 +133,7 @@ def section_field_subsection(section)
           if section[:entity_extra].present?
             possible_entities = entities(section[:entity_name], section[:entity_proxy_field])
             salient_entity = possible_entities.detect do |entity|
-              entity_label(entity, section[:entity_name]).any? { |label| label == val }
+              [entity_label(entity, section[:entity_name])].any? { |label| label == val }
```
rwd: entity_label *may* return a string but will more often return an array, so this needs to be `[entity_label(entity, section[:entity_name])].flatten.any? { |label| label == val }`

kgish: Okay, makes sense.

PR #939, opened 2017-07-25
Entity page meta tags

kgish:
Please review.

Note that thumbnail is either `src` or `full` however `full` might not exist meaning a blank image while `src` works and might be much larger.

### DIFF
--- a/app/controllers/entities_controller.rb
+++ b/app/controllers/entities_controller.rb
@@ -16,6 +16,11 @@ class EntitiesController < ApplicationController
       @entity = Europeana::API.
                 entity.fetch(entities_api_fetch_params(params[:type], params[:namespace], params[:identifier]))
       @items_by_query = build_query_items_by(params)
+      @entity_params = {
+        type: params[:type],
+        namespace: params[:namespace],
+        identifier: params[:identifier]
+      }
     end
 
     respond_to do |format|
--- a/app/controllers/entities_controller.rb
+++ b/app/controllers/entities_controller.rb
@@ -16,11 +16,6 @@ class EntitiesController < ApplicationController
       @entity = Europeana::API.
                 entity.fetch(entities_api_fetch_params(params[:type], params[:namespace], params[:identifier]))
       @items_by_query = build_query_items_by(params)
-      @entity_params = {
-        type: params[:type],
-        namespace: params[:namespace],
-        identifier: params[:identifier]
-      }
     end
 
     respond_to do |format|
--- a/app/views/entities/show.rb
+++ b/app/views/entities/show.rb
@@ -51,7 +51,7 @@ module Entities
           { meta_property: 'og:description', content: head_meta_description },
           { meta_property: 'og:url', content: request.original_url }
         ]
-        head_meta << { meta_property: 'og:title', content: title } unless title.nil?
+        head_meta << { meta_property: 'og:title', content: title }
         head_meta << { meta_property: 'og:image', content: og_image } unless og_image.nil?
         head_meta + super
       end
--- a/app/views/entities/show.rb
+++ b/app/views/entities/show.rb
@@ -64,8 +64,10 @@ module Entities
     end
 
     def og_image
-      thumbnail = entity_thumbnail
-      thumbnail ? thumbnail[:src] : nil
+      mustache[:og_image] ||= begin
+        thumbnail = entity_thumbnail
+        thumbnail ? thumbnail[:src] : nil
+      end
     end
 
     def content
--- a/app/views/entities/show.rb
+++ b/app/views/entities/show.rb
@@ -45,7 +45,7 @@ module Entities
         description = truncate(entity_description, length: 350, separator: ' ')
         description = description.strip! || description
         thumbnail = entity_thumbnail
-        image = entity[:src] || entity[:full] || nil
+        image = thumbnail ? (thumbnail[:src] || thumbnail[:full] || nil) : nil
         url = entities_fetch_url(@entity_params)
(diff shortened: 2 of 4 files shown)


kgish:
@lbiedinger : Please review (and merge).

app/views/entities/show.rb
```diff
@@ -38,24 +39,28 @@ def include_nav_searchbar
...
-    #     head_meta << { meta_property: 'og:title', content: title } unless title.nil?
-    #     head_meta << { meta_property: 'og:image', content: @landing_page.og_image } unless @landing_page.og_image.blank?
-    #     head_meta + super
-    #   end
-    # end
+    def head_meta
+      mustache[:head_meta] ||= begin
+        title = page_title
+        description = truncate(entity_description, length: 350, separator: ' ')
+        description = description.strip! || description
```
lbiedinger: rather than do this here at the top of the method, can we implement a new method for head_meta_description that also assigns the head_meta_description into the mustache data structure same as in views/collections/show.rb ?

app/views/entities/show.rb
```diff
@@ -38,24 +39,28 @@ def include_nav_searchbar
...
+        thumbnail = entity_thumbnail
+        image = thumbnail ? (thumbnail[:src] || thumbnail[:full] || nil) : nil
```
lbiedinger: let's move the image retrieval logic into a method too, something like 
```
def og_image
  thumbnail = entity_thumbnail
  image = thumbnail ? (thumbnail[:src] || thumbnail[:full] || nil) : nil
end
```
Also I think the entity_thumbnail method could use some refactoring to store to mustache[:entity_thumbnail] so then you don't even need to assign it to 'thumbnail' here.
if you do refactor the entity_thumbnail method, why do we need to assign full separately to src? I don't think it's being used.

spec/views/entities/show_spec.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+RSpec.describe 'entities/show.html.mustache' do
+  it 'should have meta description', skip: true do
+    description = 'hello'
```
lbiedinger: where is hello coming from?

app/controllers/entities_controller.rb
```diff
@@ -16,6 +16,11 @@ def show
       @entity = Europeana::API.
                 entity.fetch(entities_api_fetch_params(params[:type], params[:namespace], params[:identifier]))
       @items_by_query = build_query_items_by(params)
+      @entity_params = {
```
rwd: This is unnecessary. Just use `request.original_url` in the view.

app/views/entities/show.rb
kgish: Okay.

app/views/entities/show.rb
kgish: Okay

spec/views/entities/show_spec.rb
kgish: Cruft removed.

app/controllers/entities_controller.rb
kgish: Okay

app/views/entities/show.rb
```diff
@@ -42,26 +42,30 @@ def include_nav_searchbar
...
         ]
         head_meta << { meta_property: 'og:title', content: title } unless title.nil?
-        head_meta << { meta_property: 'og:image', content: image } unless image.nil?
+        head_meta << { meta_property: 'og:image', content: og_image } unless og_image.nil?
         head_meta + super
       end
     end
 
+    def head_meta_description
+      truncate(entity_description, length: 350, separator: ' ')
```
lbiedinger: was there no need for the ```.strip``` in the end?
Also doing this will stop running the truncation twice:
```
mustache[:head_meta_description] =|| begin
  truncate(entity_description, length: 350, separator: ' ')
end
```

kgish: Strip not needed, s/=\|\|/||=/

lbiedinger: yes ||=

app/views/entities/show.rb
```diff
@@ -38,24 +39,34 @@ def include_nav_searchbar
...
+    end
+
+    def head_meta_description
+      mustache[:head_meta_description] ||= begin
+        truncate(entity_description, length: 350, separator: ' ')
+      end
+    end
+
+    def og_image
+      thumbnail = entity_thumbnail
```
lbiedinger: actually because this is being called twice, once to check if it's nil, and once to assign it into the content of the meta tag, this should be wrapped in a block too.
```
mustache[:og_image] ||= begin
  ...
end
```

kgish: Okay.

app/views/entities/show.rb
```diff
@@ -38,24 +39,36 @@ def include_nav_searchbar
...
+        head_meta = [
+          { meta_name: 'description', content: head_meta_description },
+          { meta_property: 'fb:appid', content: '185778248173748' },
+          { meta_name: 'twitter:card', content: 'summary' },
+          { meta_name: 'twitter:site', content: '@EuropeanaEU' },
+          { meta_property: 'og:sitename', content: title },
+          { meta_property: 'og:description', content: head_meta_description },
+          { meta_property: 'og:url', content: request.original_url }
+        ]
+        head_meta << { meta_property: 'og:title', content: title } unless title.nil?
```
rwd: When would `title` ever be `nil`?

kgish: You're right, will remove.

PR #944, opened 2017-07-26
Entity social share url.

kgish:


app/views/concerns/entity_displaying_view.rb
```diff
@@ -50,7 +50,7 @@ def entity_thumbnail
 
   def entity_social_share
     {
-      url: 'this page url for share links',
+      url: 'Images from ' + entity_name + ' on Europeana Collections',
```
rwd: This is still not a URL.

PR #946, opened 2017-07-26
Feat/2303 blog caches

lbiedinger:
Change feed retrieval schedule to hourly.
Invalidate page cache when related feeds were updated.
Update the logic for invalidating caches to only happen when feeds actually changed, based on last_modified date.

### DIFF
--- a/app/jobs/cache/feed_job.rb
+++ b/app/jobs/cache/feed_job.rb
@@ -8,7 +8,7 @@ module Cache
       end
     }
 
-    queue_as :default
+    queue_as :cache
 
     def perform(url, download_media = false)
       @url = url
@@ -19,9 +19,10 @@ module Cache
         end
       end
 
-      cached_feed = Rails.cache.fetch("feed/#{url}")
+      feed_cache_key = Feed.find_by_url(url).cache_key
+      cached_feed = Rails.cache.fetch(feed_cache_key)
       if cached_feed.blank? || cached_feed.last_modified != @feed.last_modified
-        Rails.cache.write("feed/#{url}", @feed)
+        Rails.cache.write(feed_cache_key, @feed)
         @updated = true
       end
 
@@ -41,8 +42,10 @@ module Cache
     if @updated
       Cache::Expire::GlobalNavJob.perform_later if NavigableView.feeds_included_in_nav_urls.include?(@url)
       Page.joins(:feeds).where('feed.url' => @url).each do |page|
-        Cache::Expire::PageJob.perform_later(page)
+        Cache::Expire::PageJob.perform_later(page.id)
       end
     end
   end
+
+
 end
--- a/app/jobs/cache/feed_job.rb
+++ b/app/jobs/cache/feed_job.rb
@@ -11,33 +11,37 @@ module Cache
     queue_as :cache
 
     def perform(url, download_media = false)
-      feed = begin
+      @url = url
+      @download_media = download_media
+
+      @feed = begin
         benchmark("[Feedjira] #{url}", level: :info) do
-          Feedjira::Feed.fetch_and_parse(url)
+          Feedjira::Feed.fetch_and_parse(@url)
         end
       end
 
-      feed_cache_key = Feed.find_by_url(url).cache_key
+      feed_cache_key = Feed.find_by_url(@url).cache_key
       cached_feed = Rails.cache.fetch(feed_cache_key)
-      if cached_feed.blank? || cached_feed.last_modified != feed.last_modified
-        Rails.cache.write(feed_cache_key, feed)
-        updated = true
+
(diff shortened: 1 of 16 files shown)


app/jobs/cache/feed_job.rb
```diff
@@ -19,7 +19,11 @@ def perform(url, download_media = false)
         end
       end
 
-      Rails.cache.write("feed/#{url}", @feed)
+      cached_feed = Rails.cache.fetch("feed/#{url}")
+      if cached_feed.blank? || cached_feed.last_modified != @feed.last_modified
```
rwd: On reflection, I think we should let `Feedjira` do the work for us here in detecting whether the feed has been updated, i.e. something like:
```ruby
# @see https://github.com/feedjira/feedjira/blob/v2.1.0/lib/feedjira/feed_utilities.rb#L53
cached_feed.update_from_feed(@feed)
if cached_feed.updated?
  # ...
end
```

app/jobs/cache/feed_job.rb
```diff
@@ -19,7 +19,11 @@ def perform(url, download_media = false)
         end
       end
 
-      Rails.cache.write("feed/#{url}", @feed)
+      cached_feed = Rails.cache.fetch("feed/#{url}")
```
rwd: We have "feed/#{url}" all over the place... could we move that into a new method, `Feed#cache_key`?

spec/jobs/cache/expire/page_job_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      cache_key(page.cache_key, locale: 'en', user_role: 'guest')
+    end
+    let(:page_cache_content) { 'cached page' }
+
+    before do
+      Rails.cache.write(page_cache_key, page_cache_content)
+    end
+
+    it 'expires the cache entry' do
+      expect { subject.perform(page) }.to change { Rails.cache.fetch(page_cache_key) }.from(page_cache_content).to(nil)
```
rwd: Shouldn't `subject.perform(page)` be `subject.perform(page.id)`?

app/jobs/cache/feed_job.rb
```diff
@@ -32,7 +36,13 @@ def perform(url, download_media = false)
 
   # Global nav uses some feeds, and is cached so needs to be expired when those
   # feeds are updated.
+  # Cached pages need to be expired should they be using the updated feed.
   def after_perform
-    Cache::Expire::GlobalNavJob.perform_later if NavigableView.feeds_included_in_nav_urls.include?(@url)
+    if @updated
+      Cache::Expire::GlobalNavJob.perform_later if NavigableView.feeds_included_in_nav_urls.include?(@url)
+      Page.joins(:feeds).where('feed.url' => @url).each do |page|
+        Cache::Expire::PageJob.perform_later(page)
```
rwd: Shouldn't the argument to `#perform_later` be `page.id`?

app/jobs/cache/feed_job.rb
```diff
@@ -41,7 +42,7 @@ def after_perform
     if @updated
```
rwd: 1. Could we change this to `return unless @updated`?
2. I reckon we should also only enqueue the `DownloadRemoteMediaObjectJob` jobs if the feed was updated, so the code to handle that from the end of `#perform` could be moved into the `after_perform` callback.

spec/models/feed_spec.rb
```diff
@@ -69,4 +69,11 @@
       expect { subject.send(:queue_retrieval) }.to change { feed_jobs.call.count }.by(1)
     end
   end
+
+  describe '#cache_key' do
+    let(:feed) { feeds(:fashion_tumblr) }
+    it 'should append "feed/" to the url' do
```
rwd: `s/append/prepend`

app/jobs/cache/feed_job.rb
```diff
@@ -11,38 +11,34 @@ class FeedJob < ApplicationJob
...
+        Rails.cache.write(feed_cache_key, feed)
+        updated = true
       end
+      after_perform(feed, url, download_media) if updated
+    end
 
+    # Global nav uses some feeds, and is cached so needs to be expired when those
+    # feeds are updated.
+    # Cached pages need to be expired should they be using the updated feed.
+    def after_perform(feed, url, download_media)
```
rwd: This looks to be a (my?) mistake in the original writing of the `after_perform` callback.

If you look at http://guides.rubyonrails.org/v4.2/active_job_basics.html#callbacks, it only shows callbacks implemented as executions of class methods, not as definitions of instance methods.

Could we therefore rewrite this to use the (proper) form:
```ruby
module Cache
  class FeedJob < ApplicationJob
    # ...

    after_perform do |job|
      # enqueue download media jobs
      # enqueue cache expiration jobs
    end
  end
end
```

PR #947, opened 2017-07-26
Language-aware entity suggestions

rwd:
Re: https://app.assembla.com/spaces/europeana-npc/tickets/2307-entities-autosuggest--front-end--implementation-of-the-autosuggest-box-design-on-h/details

### DIFF
--- a/app/views/concerns/searchable_view.rb
+++ b/app/views/concerns/searchable_view.rb
@@ -40,23 +40,23 @@ module SearchableView
     return [] if keys.blank?
 
     keys.map do |param_key|
-      input_search_value_one_param(param_key)
+      input_search_value_from_param(param_key)
     end.flatten.compact
   end
 
-  def input_search_value_one_param(local_key, local_params = params, params_key = local_key)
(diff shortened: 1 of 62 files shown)


app/views/concerns/searchable_view.rb
```diff
@@ -43,42 +39,49 @@ def input_search
...
-  # @param key [Symbol] parameter key
-  # @param value value of the parameter
-  # @return [String] text to display
-  def input_search_param_value(key, value)
-    case key
-    when :mlt
-      response, doc = controller.fetch(value)
-      item = render_index_field_value(doc, ['dcTitleLangAware', 'title'])
-      t('site.search.similar.prefix', mlt_item: item)
+  def input_search_value_one_param(local_key, local_params = params, params_key = local_key)
```
lbiedinger: I think it might be clearer to rename this `input_search_value_from_param(_key)` instead of `input_search_value_one_param` 
If this is renamed `input_search_value_one_param_field` should also be renamed to `input_search_value_from_param_field` too.

PR #949, opened 2017-07-27
Feat/2254 consistent and clear titles

kgish:


kgish:
@lbiedinger assigned to you, please finish off exhibitions and when ready review. Feel free to resolve conflicts and merge.

app/views/collections/show.rb
```diff
@@ -30,10 +30,8 @@ def head_meta
       end
     end
 
-    def page_title
-      mustache[:page_title] ||= begin
-        @landing_page.title || @collection.title
-      end
+    def page_content_heading
+      @collection.title
```
lbiedinger: we should probably still use the landing_page title as a preference since this gives editors control over the title without changing the collection name, which only administrators can do.
also by removing the mutache[:page_title] this is being called twice now. you could assign it into mustache[:page_content_heading]

app/views/portal/show.rb
```diff
@@ -36,11 +36,8 @@ def head_meta
       end
     end
 
-    def page_title
-      mustache[:page_title] ||= begin
-        title = [display_title, creator_title]
-        CGI.unescapeHTML(title.compact.join(' | ')) + ' - Europeana'
-      end
+    def page_content_heading
+      t('site.search.page-title', default: 'Search Results')
```
lbiedinger: I don't think we want to replace the show pages with this, 'Search Results' should only be on search results(plural/index) pages not search result(singular/show) pages

app/views/collections/show.rb
kgish: Need to verify with @ashmarriotteuropeana .

app/views/portal/show.rb
kgish: Makes sense, nice catch.

app/views/collections/show.rb
kgish: Verified and specs updated.

lbiedinger: latest word from ash is to go with the collection title so this is fine as is

app/views/collections/show.rb
```diff
@@ -31,7 +31,9 @@ def head_meta
     end
 
     def page_content_heading
-      @collection.title
+      mustache[:page_content_heading] ||= begin
+        (@landing_page_title ? @landing_page_title + ' - ' : '') + @collection.title
```
lbiedinger: We don't want ```@landing_page_title``` because they always start with 'Europeana', ```@collection.title``` should be unique though.

app/views/portal/show.rb
```diff
@@ -36,13 +36,6 @@ def head_meta
       end
     end
 
-    def page_title
-      mustache[:page_title] ||= begin
-        title = [display_title, creator_title]
-        CGI.unescapeHTML(title.compact.join(' | ')) + ' - Europeana'
```
lbiedinger: This logic needs to still be kept around, otherwise all search result pages just have ' - Europeana Collections' as their title.
Following the convention the pages should be titled: 'Record Title  - Europeana Collections', so you may want to use  page_content_heading instead of page_title here.

app/views/collections/show.rb
kgish: Sorry, misunderstanding on my part.

kgish: Done, no more need for memoization.

PR #955, opened 2017-08-02
Feat/2329 birth and death dates

kgish:


spec/helpers/i18n_helper_spec.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+RSpec.describe I18nHelper do
+  describe '#date_eras_gregorian' do
+    context 'before christ' do
+      it 'converts gregorian before christ to correct format' do
+        ['-1234', ' -1234', '-1234 ', ' -1234 ', '- 1234', ' - 1234', '- 1234 ', ' - 1234 '].each do |date|
+          formatted_date = helper.date_eras_gregorian(date)
+          expect(formatted_date).to eq('1234 BCE'), "date = '#{date}' has failed"
```
rwd: The default RSpec failure message, e.g. 
```
       expected: "66 CE"
            got: "066"
```
is more helpful than this custom one, so I suggest removing the latter.

spec/helpers/i18n_helper_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+
+      it 'strips off leading zeroes' do
+        expect(helper.date_eras_gregorian('-0033')).to eq('33 BCE')
+      end
+    end
+
+    context 'after christ' do
+      it 'converts gregorian after christ to correct format' do
+        ['+12', ' +12', '+12 ', ' +12 ', '+ 12', ' + 12', '+ 12 ', ' + 12 '].each do |date|
```
rwd: This should also pass for the value `'12'`

spec/helpers/i18n_helper_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+          expect(formatted_date).to eq('12 CE'), "date = '#{date}' has failed"
+        end
+      end
+
+      it 'converts gregorian after christ yyyy-mm-dd to correct format' do
+        expect(helper.date_eras_gregorian('+56-01-01')).to eq('56-01-01 CE')
+      end
+
+      it 'strips off leading zeroes' do
+        expect(helper.date_eras_gregorian('+066')).to eq('66 CE')
```
rwd: This should also pass for the value `'066'`

app/views/concerns/entity_displaying_view.rb
```diff
@@ -217,7 +206,7 @@ def entity_death
 
   def entity_date(date)
     # Just grab the first date in the array if present.
-    date && date.is_a?(Array) && date.length && date.first.is_a?(String) ? date.first : nil
+    date && date.is_a?(Array) && date.length && date.first.is_a?(String) ? date_eras_gregorian(date.first) : nil
```
rwd: We should not just use the first string value of an array but scan the array for the first string value having a format supported by `#date_eras_gregorian`, use that if present, else use the first value in the array.

app/helpers/i18n_helper.rb
```diff
@@ -32,4 +34,17 @@ def language_options_for_select
     localised.sort_by!(&:first)
     options_for_select(localised, I18n.locale.to_s)
   end
+
+  def date_eras_gregorian(date)
+    return date unless date.is_a?(String)
+    m = date.match(/^\s*(\+|\-)\s*0*(.*)$/)
```
rwd: The final substring should only match digits, i.e. `(\d+)`

spec/helpers/i18n_helper_spec.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+RSpec.describe I18nHelper do
+  describe '#date_eras_gregorian' do
+    context 'before christ' do
+      it 'converts gregorian before christ to correct format' do
```
rwd: Let's refer to "CE" or "BCE" instead of "before christ" and "after christ" in the specs.

app/helpers/i18n_helper.rb
kgish: The problem with that is that sometime you get values like `-123-12-15` or `+1482-10-12`

spec/helpers/i18n_helper_spec.rb
kgish: Done

spec/helpers/i18n_helper_spec.rb
kgish: Done.

app/views/concerns/entity_displaying_view.rb
kgish: I was told by Hugo that we should always use the "longest" date in the array. See: https://app.assembla.com/spaces/europeana-npc/tickets/realtime_list?ticket=2335

spec/helpers/i18n_helper_spec.rb
kgish: I see you added '12' to the list, but according to the specs it think that it should not be converted since it doesn't have a '+' or '-', we can discuss later.

spec/helpers/i18n_helper_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      end
+
+      it 'converts years < 1000 to correct gregorian CE format' do
+        %w(6 66 666).each do |date|
+          expect(helper.date_eras_gregorian(date)).to eq(date + ' CE')
+        end
+      end
+
+      it 'converts gregorian CE yyyy-mm-dd to correct format' do
+        expect(helper.date_eras_gregorian('+56-01-01')).to eq('56-01-01 CE')
```
rwd: The argument to the helper here should have a four digit year, not two digit.

spec/helpers/i18n_helper_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      it 'strips off leading zeroes' do
+        %w(+006 006).each do |date|
+          expect(helper.date_eras_gregorian(date)).to eq('6 CE')
+        end
+      end
+    end
+
+    context 'normal dates' do
+      it 'leaves normal dates untouched' do
```
rwd: Not quite untouched as they are stripped.

app/helpers/i18n_helper.rb
```diff
@@ -32,4 +34,38 @@ def language_options_for_select
...
+    # Otherwise do nothing.
+    date.strip
+  end
+
+  private
+
+  def build_date(year, prefix)
+    year = strip_leading_zeroes(year)
+    key = 'global.date.eras.gregorian.' + (prefix == '+' ? 'current' : 'before')
+    default = year + ' ' + (prefix == '+' ? 'CE' : 'BCE')
```
rwd: This should not be necessary given that the English will always be present in localisations.

app/helpers/i18n_helper.rb
```diff
@@ -32,4 +34,38 @@ def language_options_for_select
...
+
+    # Otherwise do nothing.
+    date.strip
+  end
+
+  private
+
+  def build_date(year, prefix)
+    year = strip_leading_zeroes(year)
+    key = 'global.date.eras.gregorian.' + (prefix == '+' ? 'current' : 'before')
```
rwd: Instead of this, use:
```ruby
key = prefix == '+' ? 'current' : 'before'
t(key, scope: 'global.date.eras.gregorian', year: year)
```

app/helpers/i18n_helper.rb
```diff
@@ -32,4 +34,38 @@ def language_options_for_select
...
+    m = date.match(%r{^\s*(\+|\-)\s*(\d[\d\-\/]*)\s*$})
+    return build_date(m[2], m[1]) if m
+
+    # Match abnormal dates, e.g. with format 'c. AD 46' etc.
+    ['a. de C.', 'c. AD'].each do |abbrev|
+      m = date.match(%r{^\s*#{abbrev}\s*(\d[\d\-\/])\s*$})
+      return build_date(m[1], '+') if m
+    end
+
+    # Otherwise do nothing.
```
rwd: Not quite nothing: it's stripped.

spec/helpers/i18n_helper_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        %w(6 66 666).each do |date|
+          expect(helper.date_eras_gregorian(date)).to eq(date + ' CE')
+        end
+      end
+
+      it 'converts gregorian CE yyyy-mm-dd to correct format' do
+        expect(helper.date_eras_gregorian('+56-01-01')).to eq('56-01-01 CE')
+      end
+
+      it 'converts gregorian CE yyyy/mm/dd to correct format' do
```
rwd: This format is never used in Entity Collection dates, nor is it a common convention, so let's drop it from specs and helper.

app/helpers/i18n_helper.rb
kgish: Done.

spec/helpers/i18n_helper_spec.rb
kgish: Removed.

app/helpers/i18n_helper.rb
kgish: Done.

app/helpers/i18n_helper.rb
kgish: Done.

spec/helpers/i18n_helper_spec.rb
kgish: Done.

app/views/concerns/entity_displaying_view.rb
lbiedinger: this is just taking the first date still, should we run date_eras_gregorian() against each date and return the first one that it matches?

PR #956, opened 2017-08-03
Friendly entity page URLs

rwd:
Re: https://app.assembla.com/spaces/europeana-npc/tickets/2316

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -60,11 +60,11 @@
 
 ## Change the URL of the Europeana Search/Record REST API from the default,
 ## production, URL <https://www.europeana.eu/api>. No trailing slash.
-# EUROPEANA_API_URL=http://api.example.com/
+# EUROPEANA_API_URL=http://other-search.example.com/api
 
 ## Change the URL of the Europeana Entities REST API from the default,
 ## production, URL <https://www.europeana.eu/api>. No trailing slash.
-# EUROPEANA_ENTITIES_API_URL=http://entity-api.example.com/
+# EUROPEANA_ENTITIES_API_URL=http://other-entity.example.com/api
 
 ## Change the URL of the proxy service for media downloads from the default,
 ## production, URL <http://proxy.europeana.eu>. No trailing slash.
--- a/.env.example
+++ b/.env.example
@@ -3,8 +3,11 @@
 ## URL of the PostgreSQL database.
 # DATABASE_URL=postgres://user:pass@host/db
 
-## Your Europeana REST API key. Request one at http://labs.europeana.eu/api/registration
+## Your Europeana Search/Record API key. Request one at http://labs.europeana.eu/api/registration
 # EUROPEANA_API_KEY=
+
+## Your Europeana Entities API key.
+# EUROPEANA_ENTITIES_API_KEY=
 
 ## Session secret key. Generate one with `bundle exec rake secret`
 # SECRET_KEY_BASE=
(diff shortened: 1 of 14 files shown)


.env.example
```diff
@@ -55,9 +58,13 @@
 ## Send emails reporting application errors to this email address.
 # ERROR_REPORT_MAIL_TO=developers@example.com
 
-## Change the URL of the Europeana REST API from the default, production, URL
-## <https://www.europeana.eu/api/v2>. No trailing slash.
-# EUROPEANA_API_URL=http://api.example.com/v2
+## Change the URL of the Europeana Search/Record REST API from the default,
+## production, URL <https://www.europeana.eu/api>. No trailing slash.
+# EUROPEANA_API_URL=http://api.example.com/
```
lbiedinger: the comment says "not trailing slash", but the example has a trailing slash

.env.example
```diff
@@ -55,9 +58,13 @@
...
+
+## Change the URL of the Europeana Entities REST API from the default,
+## production, URL <https://www.europeana.eu/api>. No trailing slash.
+# EUROPEANA_ENTITIES_API_URL=http://entity-api.example.com/
```
lbiedinger: again, comment and example at odds.

config/initializers/env.rb
```diff
@@ -14,7 +14,11 @@
                                 else
                                   ''
                                 end
-  europeana[:entities_api_key] = ENV['EUROPEANA_ENTITIES_API_KEY'] || Europeana::API.key
+
+  europeana[:entities] = OpenStruct.new(
+    api_key: ENV['EUROPEANA_ENTITIES_API_KEY'] || Europeana::API.key,
+    api_url: ENV['EUROPEANA_ENTITIES_API_URL'] || Europeana::API.url
```
lbiedinger: This is causing problems for some of the specs that with the setup I have/had my .env files in.
I was setting `EUROPEANA_API_URL` to the acceptance api in .env and not setting the entities API url at all.
I guess, when running rspec, `Europeana::API.url` hasn't been set to the ENV value at the point when this initializer is ran, because the tests stubbed all the accpetance requests but when executing fixtures it was making requests to the default API url from the API ruby gem.

it's more robust if we do this:
```api_url: ENV['EUROPEANA_ENTITIES_API_URL'] || ENV['EUROPEANA_API_URL'] || Europeana::API.url```

Comment (lbiedinger) [inline config/initializers/env.rb]: probably should do the same for the entites api key too

File: app/helpers/entities_helper.rb [hunk 4]
```diff
@@ -0,0 +1,44 @@
...
+  ##
+  # Constructs a URL slug for an entity page from the English prefLabel
+  #
+  # If there is no English prefLabel, the entity page has no URL slug, and
+  # the return value will be `nil`.
+  #
+  # @param entity [HashWithIndifferentAccess] Response from `Europeana::API.entity.fetch`
+  # @return [String] URL slug
+  def entity_url_slug(entity)
+    return nil unless entity.key?(:prefLabel) && entity[:prefLabel].key?(:en)
```
lbiedinger: Can we be sure that this will always be :en for single entity fetch responses? The auto suggest was always returning quite a few responses with an empty key where there's only one default prefLabel.

lbiedinger: This is a question, since I'm not able to find examples that would suggest that always relying on en isn't going to be enough.

spec/controllers/entities_controller_spec.rb
```diff
@@ -62,22 +69,55 @@
         Rails.application.config.x.enable.entity_page = true
       end
 
-      it 'returns http success' do
-        get :show, locale: 'en', type: 'agent', namespace: 'base', identifier: '1234'
+      context 'without slug in URL' do
+        it 'redirects to URL with slug' do
+          allow(subject).to receive(:authorize!) { true }
```
lbiedinger: anywhere within these entity controller specs this authorize method is stubbed, it's 'legacy' code from when we were hiding entity pages via authorization.
These aren't being used now that we enable entity pages via the env variable.
Could you remove the 4 instances of this?

app/helpers/entities_helper.rb
rwd: We can not be sure that there will be an English prefLabel, but if there is not, we should not use the prefLabel with undefined language (key `""`) as all of our other URL paths are in English. In those cases, no textual slug will be used.

config/initializers/env.rb
rwd: Good point. I think the better solution is just not to set them if the env vars are not present.

PR #964, opened 2017-08-09
Optimize search view

rwd:
Re: https://europeanadev.assembla.com/spaces/europeana-npc/tickets/858

1. Adds [StackProf](https://github.com/tmm1/stackprof) to the bundle, and adds a new environment named "profiling". To profile the application, run `RAILS_ENV=profiling bundle exec puma -C config/puma.rb`, execute some requests, then `RAILS_ENV=profiling bundle exec stackprof tmp/stackprof-*.dump`
2. Updates Ruby to 2.4.1. Compared to 2.3.1, profiling show less wall time spent in `CGI.escape`, with a ~20% performance improvement to the search results page which contains many links and hence makes many calls to that method (down in the depths of the Rails framework).
3. Updates Rails and Puma gems, and the rest of the bundle. Primarily for Ruby 2.4 support, but also just for housekeeping. Unlocks gems that were locked pending upstream fixes which have been released.
4. Optimises `EDM::Rights.from_api_query` by storing a map of URLs to rights data.
5. Fixes an incorrect job name in the model observer `NavCacheInvalidationObserver`


config/environments/profiling.rb
```diff
@@ -0,0 +1,84 @@
...
+  # `config.assets.precompile` and `config.assets.version` have moved to config/initializers/assets.rb
+
+  # Specifies the header that your server uses for sending files.
+  # config.action_dispatch.x_sendfile_header = 'X-Sendfile' # for Apache
+  # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
+
+  # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
+  # config.force_ssl = true
+
+  config.log_level = :debug
```
lbiedinger: in production this is set to `:info` is this deliberate?

rwd: Yes, as profiling would (for now) only be run on a dev installation of the app where full logging would be useful.

PR #966, opened 2017-08-14
Configurable additional locales

rwd:
Re: https://europeanadev.assembla.com/spaces/europeana-npc/tickets/2338

* Additional locales can be specified in the env var `I18N_ADDITIONAL_LOCALES` (see .env.example)
* Language name lookups are moved to the I18n scope `global.languages`

### DIFF
--- a/config/initializers/env.rb
+++ b/config/initializers/env.rb
@@ -40,11 +40,3 @@ Rails.application.config.x.enable = OpenStruct.new(
   search_form_autocomplete: ENV['ENABLE_SEARCH_FORM_AUTOCOMPLETE'],
   entity_page: ENV['ENABLE_ENTITY_PAGE']
 )
-
-# Add locales to available locales
-if ENV['I18N_ADDITIONAL_LOCALES']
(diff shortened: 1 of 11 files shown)


config/initializers/env.rb
```diff
@@ -40,3 +40,11 @@
   search_form_autocomplete: ENV['ENABLE_SEARCH_FORM_AUTOCOMPLETE'],
   entity_page: ENV['ENABLE_ENTITY_PAGE']
 )
+
+# Add locales to available locales
+if ENV['I18N_ADDITIONAL_LOCALES']
+  ENV['I18N_ADDITIONAL_LOCALES'].split(',').each do |locale|
+    Rails.application.config.i18n.available_locales << locale.to_sym
+  end
+  Rails.application.config.i18n.available_locales.sort!
```
lbiedinger: This might be a good place to remove duplicates, if they're part of the standard locales and added again via I18N_ADDITIONAL_LOCALES, or added multiple times through I18N_ADDITIONAL_LOCALES

PR #970, opened 2017-08-16
Feat/2335 most accurate/longest date should be shown

kgish:
New i18n helpers `date_most_accurate` and `date_score`, all tests green.


app/views/concerns/entity_displaying_view.rb
```diff
@@ -205,14 +205,7 @@ def entity_death
...
 
   def entity_date(dates)
-    return nil unless dates.present? && dates.is_a?(Array)
-
-    dates.each do |date|
-      gregorian = date_eras_gregorian(date)
-      return gregorian unless gregorian.nil?
-    end
-
-    dates.first.strip
```
lbiedinger: keep this as a fallback here

app/views/concerns/entity_displaying_view.rb
```diff
@@ -205,14 +205,7 @@ def entity_death
...
+    date_most_accurate(dates)
```
lbiedinger: rather than calling it on the dates as returned, let's call it on the dates formated with `date_eras_gregorian` and call compact on that array.

app/helpers/i18n_helper.rb
```diff
@@ -47,6 +47,40 @@ def date_eras_gregorian(date)
     nil
   end
 
+  # Return the most accurate date from an array of dates using scoring algorithm below
+  def date_most_accurate(dates)
+    return nil unless dates.present? && (dates.is_a?(Array) || dates.is_a?(String))
+    dates = [dates] if dates.is_a?(String)
+    most_accurate = nil
```
lbiedinger: assign the first gregorian date here already to avoid the extra conditional block in the each loop

app/views/concerns/entity_displaying_view.rb
kgish: Okay.

app/views/concerns/entity_displaying_view.rb
kgish: Okay.

app/views/concerns/entity_displaying_view.rb
```diff
@@ -205,14 +205,7 @@ def entity_death
   end
 
   def entity_date(dates)
-    return nil unless dates.present? && dates.is_a?(Array)
```
kgish: Okay, also check `return dates if dates.is_a?(String)`

app/helpers/i18n_helper.rb
```diff
@@ -47,6 +47,40 @@ def date_eras_gregorian(date)
     nil
   end
 
+  # Return the most accurate date from an array of dates using scoring algorithm below
+  def date_most_accurate(dates)
+    return nil unless dates.present? && (dates.is_a?(Array) || dates.is_a?(String))
+    dates = [dates] if dates.is_a?(String)
```
kgish: Done.

app/helpers/i18n_helper.rb
kgish: Done.

app/views/concerns/entity_displaying_view.rb
```diff
@@ -205,7 +205,9 @@ def entity_death
   end
 
   def entity_date(dates)
-    date_most_accurate(dates)
+    return dates if dates.is_a?(String)
```
lbiedinger: we still need to apply `date_eras_gregorian(dates)` and possibly `.strip` even if passed a string.
rather than returning dates when entity_date is passed a string, either assign dates = [dates] or remove this line altogether and leave it to the calling method to supply an array.

app/helpers/i18n_helper.rb
```diff
@@ -49,20 +49,17 @@ def date_eras_gregorian(date)
...
-    most_accurate = nil
-    max_score = 0
-    dates.map { |date| date_eras_gregorian(date) || date.strip }.each_with_index do |date, index|
-      if index.zero?
-        max_score = date_score(date)
+    return nil unless dates.present?
+    return dates if dates.is_a?(String)
+    return dates[0] if dates.length == 1
+    date_eras = dates.map { |date| date_eras_gregorian(date) || date }.compact
+    most_accurate = date_eras.shift
```
lbiedinger: you could get rid off this line altogether if we assign an initial value of -1 to max_score

kgish: Okay.

app/helpers/i18n_helper.rb
```diff
@@ -49,20 +49,17 @@ def date_eras_gregorian(date)
...
   def date_most_accurate(dates)
-    return nil unless dates.present? && (dates.is_a?(Array) || dates.is_a?(String))
-    dates = [dates] if dates.is_a?(String)
-    most_accurate = nil
-    max_score = 0
-    dates.map { |date| date_eras_gregorian(date) || date.strip }.each_with_index do |date, index|
-      if index.zero?
-        max_score = date_score(date)
+    return nil unless dates.present?
+    return dates if dates.is_a?(String)
```
kgish: This is a helper which in principle could be called from anywhere, not just from entity_date. Also, there are tests for passing nil and string, so these break as well if I remove the lines. Prefer to keep these sanity checks in place to ensure method contract is not violated.

app/helpers/i18n_helper.rb
```diff
@@ -49,20 +49,17 @@ def date_eras_gregorian(date)
...
+    return dates[0] if dates.length == 1
```
kgish: Nice catch, done.

app/views/concerns/entity_displaying_view.rb
kgish: Since this has now been moved to `date_most_accurate` it can be removed altogether.

app/helpers/i18n_helper.rb
```diff
@@ -47,6 +47,37 @@ def date_eras_gregorian(date)
     nil
   end
 
+  # Return the most accurate date from an array of dates using scoring algorithm below
+  def date_most_accurate(dates)
+    return nil unless dates.present?
+    return dates if dates.is_a?(String)
```
lbiedinger: this is still not formatting `dates` even if we were allowing `date_most_accurate` to be called with a string as a parameter, then we still need to format that string as gregorian.

kgish: You're right, will fix.

app/helpers/i18n_helper.rb
```diff
@@ -47,6 +47,39 @@ def date_eras_gregorian(date)
...
+    most_accurate
+  end
+
+  # Give the date a score based on the possible date formats:
+  # '1957' => 1 point
+  # '1957-10' => 2 points
+  # '1957-10-11' => 3 points
+  # 'CE' or 'BCE' => 1 point extra
+  # otherwise => 0
+  def date_score(date)
```
rwd: This needs to be stricter.

The following should get scores of 0:
```ruby
'1-2 cups a day'
'--01-02'
'1-2-3-4'
```

app/helpers/i18n_helper.rb
```diff
@@ -47,6 +47,39 @@ def date_eras_gregorian(date)
...
 
+  # Return the most accurate date from an array of dates using scoring algorithm below
+  def date_most_accurate(dates)
+    return nil unless dates.present?
+    if dates.is_a?(String)
+      return date_eras_gregorian(dates) || dates.strip
+    end
+    date_eras = dates.map { |date| date_eras_gregorian(date) || date }.compact
+    return date_eras[0] if date_eras.length == 1
+    max_score = -1
```
kgish: Okay.

app/helpers/i18n_helper.rb
```diff
@@ -47,6 +47,40 @@ def date_eras_gregorian(date)
...
+  #
+  # Valid date combinations of year=y, month=m and day=d:
+  # y, yy, yyy, yyyy, y-m, yy-m, yyy-m, yyyy-m, y-mm, yy-mm, yyy-mm, yyyy-mm
+  # y-m-d, yy-m-d, yyy-m-d, yyyy-m-d, y-mm-d, yy-mm-d, yyy-mm-d, yyyy-mm-d
+  # y-m-dd, yy-m-dd, yyy-m-dd, yyyy-m-dd, y-mm-dd, yy-mm-dd, yyy-mm-dd, yyyy-mm-dd
+  #
+  # $` = string before match (if non-empty and not ending in space, score => 0)
+  # $' = string after match (if non-empty and not ' BCE' or ' CE', score => 0)
+  # m[0] = matching string, m[1-3] = matching substring
+  def date_score(date)
```
rwd: The contents of this method are difficult to read, and can be rewritten to pass existing tests as:
```ruby
  def date_score(date)
    return 0 unless date.present?

    m = date.strip.match(/^(\d{1,4})(-\d{1,2})?(-\d{1,2})?( B?CE)?$/)

    m.nil? ? 0 : m.to_a.compact.length - 1
  end
```

kgish: Okay, makes sense.

app/helpers/i18n_helper.rb
```diff
@@ -47,6 +47,38 @@ def date_eras_gregorian(date)
     nil
   end
 
+  # Return the most accurate date from an array of dates using scoring algorithm below
```
lbiedinger: Since this method description says the method takes an array of dates and returns the most specific, but the method actually does more than this (it also takes a single non formatted date and formats it; as well as doing the date formatting on the dates) either the comment should be updated, or in order to keep the functionality of this method more atomic to just select the most accurate date, the single string detection/date formatting should be moved into the caller method. Tests would also need to be updated to reflect the fact that this method does NOT take a single string value.

kgish: I would rather just change the description and preserve the ability to allow a `single date` to be passed. In theory, as is the case with other elements, the API could return a single date of the form { 'date' } and we should be able to handle this case transparently.

app/helpers/i18n_helper.rb
```diff
@@ -47,6 +47,39 @@ def date_eras_gregorian(date)
     nil
   end
 
+  # Return the most accurate date from an array of dates using scoring algorithm below, if a single
+  # date string is passed, then that value is returned immediately.
+  def date_most_accurate(dates)
+    return nil unless dates.present?
+    if dates.is_a?(String)
+      return date_eras_gregorian(dates) || dates.strip
+    end
```
lbiedinger: if we're keeping this in here, I think the above 4 lines could be simplified and prevent additional processing of empty arrays by changing them to:
```
 return nil if dates.blank?
 return date_eras_gregorian(dates) || dates.strip if dates.is_a?(String)
```

kgish: Done.

PR #978, opened 2017-08-31
EDM entity model

kgish:
Re: https://europeanadev.assembla.com/spaces/europeana-npc/tickets/2369


kgish:
@rwd and/or @lbiedinger Just realized that #980 changes will have to be merged into this, otherwise it'll get lost.

PR #980, opened 2017-08-31
Feat/2346 entity description title

kgish:


kgish:
@rwd and/or @lbiedinger Just realized that this will have to be merged into  #978, otherwise it'll get lost.

rwd:
Closing this because it's based on the wrong branch. It needs to be based on release/v2.5.0, and the PR be to merge into that. I will recreate it.

PR #988, opened 2017-09-18
Feat/2362 noindex unlinked entities

lbiedinger:


app/models/concerns/edm/entity/referenced_records.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module EDM
+  module Entity
+    module ReferencedRecords
```
rwd: I'm concerned about how tight a coupling this introduces to the Blacklight interface to the Record API, *within* the entity models.

While it would be nice to have model-level associations to retrieve CHOs etc from the Record API, I do not think it can be justified at the cost of bringing Blacklight documents, config and controller instances into these models.

I'd suggest therefore that retrieval of referenced records be moved up into `EntitiesController` which then need only use `#search_results` to construct `@referenced_records` as it already has the Blacklight config, and is a controller.

app/models/concerns/edm/entity/referenced_records.rb
```diff
@@ -0,0 +1,40 @@
...
+        !referenced_records.detect { |_key, records| records[:total][:value] != 0 }
+      end
+
+      def referenced_records
+        @refereneced_records ||= {}
+        search_keys.each do |key|
+          @refereneced_records[key] ||= begin
+            return { search_results: [], total: { value: 0, formatted: '0' } } unless respond_to?(:search_query)
+            @response = Europeana::Blacklight::Response.new(repository.search(query: search_query), controller.params)
+            {
```
rwd: This is used for display purposes, so move it into the `Entities::Show` view class.

app/models/concerns/edm/entity/referenced_records.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module EDM
+  module Entity
+    module ReferencedRecords
+      include RepositoryHelper
+      extend ActiveSupport::Concern
+
+      def unreferenced?
```
rwd: Given the above comment, this would need to become a method on the view class instead of on the model.

app/controllers/entities_controller.rb
```diff
@@ -16,14 +18,21 @@ def suggest
 
   def show
     @body_cache_key = body_cache_key if entity_caching_enabled?
-    @entity = EDM::Entity.build_from_params(entity_params) unless body_cached? && entity_caching_enabled?
-
+    unless body_cached? && entity_caching_enabled?
+      @entity = EDM::Entity.build_from_params(entity_params)
+      @entity.search_keys.each { |search_key| search_results(search_key) }
```
rwd: Replace this with:
```ruby
      @search_results = @entity.search_keys.each_with_object({}) do |search_key, results|
        entity_search_params = { q: @entity.search_query(search_key) }
        (response, _documents) = search_results(entity_search_params)
        results[search_key] = response
      end
```

app/controllers/entities_controller.rb
```diff
@@ -16,14 +18,21 @@ def suggest
...
+    end
     respond_to do |format|
       format.html
       format.json { render json: @entity }
     end
   end
 
+  def search_results(key)
```
rwd: Remove this method.

app/controllers/entities_controller.rb
```diff
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 class EntitiesController < ApplicationController
   include CacheHelper
+  include RepositoryHelper
   include Europeana::EntitiesApiConsumer
+  include ActionView::Helpers::NumberHelper
```
rwd: Remove this.

app/controllers/entities_controller.rb
```diff
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 class EntitiesController < ApplicationController
   include CacheHelper
+  include RepositoryHelper
```
rwd: Remove this.

app/helpers/repository_helper.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module RepositoryHelper
```
rwd: Delete this helper.

app/views/entities/show.rb
```diff
@@ -89,23 +90,45 @@ def thumbnail
...
+    def unreferenced?
+      !referenced_records.detect { |_key, records| records[:total][:value] != 0 }
+    end
+
+    def referenced_records
+      @referenced_records ||= {}
+      @entity.search_keys.each do |key|
+        @referenced_records[key] ||= begin
+          return { search_results: [], total: { value: 0, formatted: '0' } } unless @entity.respond_to?(:search_query)
+          @response = @search_results[key]
```
rwd: Change the use of `@response` in this method to just `response` and pass that to `#document_presenter`.

PR #989, opened 2017-09-18
re: #2368 entity page caching configurable

lbiedinger:


app/controllers/entities_controller.rb
```diff
@@ -3,7 +3,7 @@ class EntitiesController < ApplicationController
   include CacheHelper
   include Europeana::EntitiesApiConsumer
 
-  attr_reader :body_cache_key
+  attr_reader :body_cache_key, :entity_caching_enabled
```
rwd: There's no use in having an attr_reader if the method is implemented in the class.

app/controllers/entities_controller.rb
```diff
@@ -42,6 +42,10 @@ def entity
     end
   end
 
+  def entity_caching_enabled
```
rwd: Let's name this `entity_caching_enabled?` (or would that be a problem for the Mustache template...?)

lbiedinger: it doesn't work well with mustache, but since it's assigning @entity_caching_enabled it's fine as long as the method is always called before the view is rendered

PR #993, opened 2017-09-21
stop appending to msg constant

lbiedinger:


app/controllers/concerns/controller_exception_handling.rb
```diff
@@ -172,7 +172,7 @@ def template_for_html_error_response_page(page:)
   ##
   # Render a simple JSON error response
   def render_json_error_response(exception:, status:)
-    msg = Rack::Utils::HTTP_STATUS_CODES[status]
+    msg = Rack::Utils::HTTP_STATUS_CODES[status].dup
```
rwd: Instead of doing this, could we add the frozen literal string magic comment to the file, which I reckon would have caught this, then revert this line and change the next to:
```ruby
     msg = "#{msg}: #{exception.message}" unless exception.message.blank?
```

app/controllers/concerns/controller_exception_handling.rb
```diff
@@ -57,8 +59,8 @@ def process(exception:, controller:, **_)
 
         def message(exception:)
           message = "\n#{exception.class} (#{exception.message}):\n"
-          message << exception.annoted_source_code.to_s if exception.respond_to?(:annoted_source_code)
-          message << '  ' << trace(exception: exception).join("\n  ")
+          message = "#{message}#{exception.annoted_source_code}" if exception.respond_to?(:annoted_source_code)
+          "#{message} #{trace(exception: exception).join("\n  ")}"
```
rwd: Could we keep separating message and trace with two space characters here, as before, for consistency in logs?

lbiedinger: good spot!

PR #997, opened 2017-09-27
Refactor field group presentation

rwd:
This branch introduces a substantial refactor of the `Document::FieldGroupPresenter` used as the presentation logic for grouped metadata fields on the core item display pages.

Key changes:
* introduction of nested `Document::FieldGroup::SectionPresenter` and `Document::FieldGroup::Section::ValuePresenter` classes with much of the code from `Document::FieldGroupPresenter` moved into them for better OO modelling and improved readability
* replaces `DocumentPresenter` with `BlacklightDocumentPresenter` *concern* to be included in presenters needing access to the `Europeana::Blacklight::DocumentPresenter` methods
* introduces a base presenter class `ApplicationPresenter`
* europeana-blacklight gem updated to [disable-lang-map-value-dereferencing branch](https://github.com/europeana/europeana-blacklight/tree/disable-lang-map-value-dereferencing) where it no longer dereferences entity URIs in `Europeana::Blacklight::Document::LangMaps#localize_lang_map` so that consumers can obtain the URI and work with it as needed
* config/record_field_groups.yml:
  * remove `entity_proxy_field` and use `fields`
  * move other `entity_` settings under `entity`
  * enable entities lookup on format and type sections (for https://europeanadev.assembla.com/spaces/europeana-npc/tickets/2374)
* rationalised placement of entities-related methods in view/presenter concerns


app/presenters/document/field_group/section/value_presenter.rb
```diff
@@ -0,0 +1,143 @@
...
+
+module Document
+  module FieldGroup
+    module Section
+      ##
+      # Presenter for one value of a metadata field
+      #
+      # TODO: should this be sub-classed? e.g. for single string values, language
+      #   maps, entity-linked values
+      class ValuePresenter < DocumentPresenter
```
lbiedinger: Sub classing might make sense for the different types of values, but more fundamentally, should we move the DocumentPreseneter logic into a concern? I don't know if it's just the semantics not quite making sense in my head, but 'DocumentPresenter' suggests to me that this is an object to Present an entire Document. 
These ValuePresenters or if we further sub-class them, EntityValuePresenters/LinkableValuePresenters/SimpleValuePresenters/whatver, do need to Present part of a Document, but don't display the whole Document.

rwd:
@lbiedinger re:
> Apart from my suggestion about making te DocumentPresenter logic come in to our other Preseneters as a concern rather than through inheritance.
I'm also not 100% convinced by the use of modules here. While they do really help in seeing the relationship/nesting between Documents/FieldGroups/Sections/Values, they don't really do anything. As in we never say include FieldGroup and the nesting doesn't exist in terms of ValuePresenter < SectionPresenter < FieldGroupPresenter < DocumentPresenter, instead it's just ValuePresenter < DocumentPresenter, which again makes me think we should be using a concern rather than inheritance.

I don't see that there is a problem here. The modules are simply used for namespacing, not as `ActiveSupport::Concerns`, so do not need to be included elsewhere in order to serve their purpose. Just as with the `Explore` module in the view class `Explore::Colours`.

app/presenters/concerns/document/field/entities.rb
```diff
@@ -4,50 +4,51 @@ module Document
...
         @entities[type] ||= {}
-        @entities[type][proxy_field] ||= entities_for(type, proxy_field)
+        @entities[type][field] ||= document_entities_for_type(type, field)
       end
 
-      # @param (see #entities)
-      def entities_for(type, proxy_field = nil)
+      # @param (see #document_entities)
+      def document_entities_for_type(type, field = nil)
         document_entities = document.fetch(type, [])
```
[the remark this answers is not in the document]
rwd: Is that a problem? As there is no method `#document_entities=`, this will always result in a local variable instantiation.

app/presenters/document/field_group/section_presenter.rb
```diff
@@ -0,0 +1,176 @@
...
+        field_contents = document_field_contents(field)
+
+        field_contents.map do |content|
+          field_entity = field_entities.detect { |fe| fe[:about] == content }
+
+          if field_entity.present?
+            [entity_label(field_entity)].flatten.map do |label|
+              value_presenter(field, label, field_entity)
+            end
+          elsif entity_fallback?
```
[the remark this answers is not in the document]
rwd: No.

app/presenters/document/field_group/section/value_presenter.rb
rwd: I agree in principle but the difficulty is that the methods for retrieving fields from documents are declared in `Blacklight::DocumentPresenter` (https://github.com/projectblacklight/blacklight/blob/v6.0.2/app/presenters/blacklight/document_presenter.rb) so wherever we use these, we need to have sub-classed that, here via `Europeana::Blacklight::DocumentPresenter` then `DocumentPresenter`.

app/presenters/concerns/document/field/entities.rb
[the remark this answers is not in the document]
lbiedinger: not a problem from a functional point of view, just when I was reading it initially it was confusing since:
`document_entities` calls `document_entities_for_type` which at the end returns `document_entities`

app/presenters/document/field_group/section/value_presenter.rb
lbiedinger: would an option be to make a concern that assigns the DocumentPresenter to an instance variable that can then be delegated to or used by our other presenters to pull the information from the document?
Along the lines of DocumentPresenterPresentingPresenter/DocumentPresenterUsingPresenter

If that was possible, the main downside would still be that our presenters for field_groups/sections/values will be dependant on the DocumentPresenter rather than being presenters for real models. Which isn't really the intend use of presenters as far as I can make out, although that's a bit of a grey area too. However this is inherent to the fact that we don't explicitly model FieldGroups/Sections/Values explicitly but rather derive these 'models' from the Document and our configuration.

app/presenters/concerns/document/field/entities.rb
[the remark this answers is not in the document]
rwd: Renamed in 92e2dce0

app/presenters/document/field_group/section/value_presenter.rb
rwd: I have removed the sub-classing of `DocumentPresenter` except where it is absolutely necessary and introduced an `ApplicationController` base class for other presenters, which for now just delegates `#t` to `I18n`.

app/presenters/document/field_group/section/value_presenter.rb
```diff
@@ -0,0 +1,145 @@
...
+          linkable_value? ? content : nil
+        end
+
+        def linkable_value?
+          return false unless content.is_a?(String)
+          content.start_with?('http://', 'https://')
+        end
+
+        def linkable_for_entity?
+          Rails.application.config.x.enable.entity_page &&
```
lbiedinger: should we remove this configuration already?

app/presenters/document/field_group/section_presenter.rb
```diff
@@ -0,0 +1,204 @@
...
+      attr_reader :document, :controller, :group, :definition
+
+      delegate :search_field, :entity, :fields, :exclude_vals, :max, :title,
+               :ga_data, :map_values, :format_date, to: :definition
+
+      # @param document [Europeana::Blacklight::Document]
+      # @param controller [ActionController::Base]
+      # @param group [Document::FieldGroupPresenter]
+      # @param definition [OpenStruct]
+      def initialize(document, controller, group, definition)
```
lbiedinger: this is fine, but is it worth wile trying to align the flow with the initializer for the FieldGroupPresenter?
So loading the value presenters into @value_presenters similarly or how the FieldGroupPresenter initializes with the definitions for it's sections.

app/presenters/document/field_group/section/value_presenter.rb
rwd: Yes, and done in bed709e0

app/presenters/document/field_group/section_presenter.rb
rwd: I am not too sure what you mean here. Could you show some code to illustrate?

lbiedinger: the FieldGroupPresenter loads in section presenters for each section at initialization:
```
    ...
    @sections = @definition[:sections].map do |section_definition|
      Document::FieldGroup::SectionPresenter.new(document, controller, self, OpenStruct.new(section_definition).freeze)
    end
end
```

the sections don't load in the value_presenters until .display is called
so for consistency in the initializer here we could do something like: 
```
@value_presenters = value_presenters
```
however the logic for determining the value presenters is a bit more complicated for the section presenters than the simple mapping we can do in the FieldgroupPresenter initializer. Therfore I was asking whether it's worth trying to move this into the initializer.

lbiedinger: or potentially move the assignment of @sections out of the initializer in the FieldGroupPresenter class.

rwd: Understood. I have moved the initialisation of the section presenters out of the field group presenter initializer, in 32cf610c20527a1fd66e626e4670eab3a02ea6e6.

PR #998, opened 2017-10-04
EUscreen tweaks

rwd:
Re: https://app.assembla.com/spaces/europeana-npc/tickets/2426

### DIFF
--- a/app/presenters/document/web_resource_presenter.rb
+++ b/app/presenters/document/web_resource_presenter.rb
@@ -206,17 +206,17 @@ module Document
     def displayable_tests
       [
         # TRUE if for edm:object and no other web resources are displayable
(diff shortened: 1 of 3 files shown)


app/presenters/document/web_resource_presenter.rb
```diff
@@ -233,18 +259,26 @@ def downloadable?
...
+        for_has_view? || for_edm_is_shown_by?
       end
     end
 
     def for_edm_object?
-      @record_presenter.edm_object == url
+      url == @record_presenter.edm_object
+    end
+
+    def for_edm_is_shown_at?
```
lbiedinger: this isn't being called, is there a dynamically constructed method call or for potential future use?

rwd: Nope, you're right, it's redundant.

PR #999, opened 2017-10-04
Feat/2425 facet range refinements

lbiedinger:


app/controllers/concerns/blacklight_config.rb
```diff
@@ -81,8 +81,9 @@ def self.collections_query_facet
       config.add_facet_field 'YEAR',
                              range: true,
                              when: ->(context) { context.within_collection? && context.current_collection.key == 'fashion' },
-                             limit: 1_000,
-                             only: ->(item) { item.value =~ /\A-?\d{1,4}\z/ }
+                             limit: 2_000,
+                             only: ->(item) { item.value =~ /\A-?\d{1,4}\z/ },
+                             format_value_as: ->(value) { value.to_i }
```
rwd: Do you think that the `format_value_as` option should be renamed to `format` to be more in keeping with the other short option names?

lbiedinger: I had originally called it format, but my IDE complained about the call to format missing arguments, since there's a private format method on the objects already. 
Now it still works, if I use format for the option, but to avoid ambiguity I gave it a more specific name. If you want I can rename to format, or format_value(s) or leave it as it is.

rwd: Understood, thanks! It's fine as it is, then.

PR #1002, opened 2017-10-05
bigger api parameter box

lbiedinger:


config/initializers/rails_admin.rb
```diff
@@ -135,7 +135,9 @@
         end
       end
       field :title
-      field :api_params
+      field :api_params, :text do
+        html_attributes rows: 15, style: 'width: 100%;'
```
rwd: Could you move this style setting into app/assets/stylesheets/rails_admin/custom/theming.scss, as:
```css
body.rails_admin .form-horizontal textarea[name="collection[api_params]"] {
  width: 100%;
}
```

PR #1004, opened 2017-10-11
Unescape webresource urls

lbiedinger:


app/presenters/document/web_resource_presenter.rb
```diff
@@ -73,7 +73,7 @@ def play_html
 
     def url
       @url ||= begin
-        url = field_value('about')
+        url = field_value('about', unescape: true)
```
rwd: Instead of doing this, which results in HTML escaping and then unescaping, we should instead use `url = document.fetch('about', nil)`

PR #1006, opened 2017-10-17
use feed-job gem

lbiedinger:


app/jobs/cache/feed_job.rb
```diff
@@ -1,42 +1,19 @@
...
     # feeds are updated.
     # Cached pages need to be expired should they be using the updated feed.
     after_perform do
-      return unless @updated
-      if @download_media
-        @feed.entries.each do |entry|
-          img_url = FeedEntryImage.new(entry).media_object_url
-          DownloadRemoteMediaObjectJob.perform_later(img_url) unless img_url.nil?
+      if @updated
+        if @download_media
```
rwd: `@download_media` should not be set by `Europeana::Feeds::FetchJob` as it does nothing there.

Scanning this repo's codebase, all uses of `Cache::FeedJob` *do* download media, so this condition can it seems just be removed.

app/models/feed.rb
```diff
@@ -11,6 +11,12 @@ class Feed < ActiveRecord::Base
 
   acts_as_url :name, url_attribute: :slug, only_when_blank: true, allow_duplicates: false
 
+  def self.exhibitions_urls
```
rwd: Does this belong here, given #1015?

app/views/application_view.rb
```diff
@@ -59,7 +59,7 @@ def head_links
   end
 
   def fb_campaigns_on
-    true
+    true unless Rails.env.test?
```
rwd: What's this?

app/jobs/cache/feed_job.rb
lbiedinger: good point, these were in here from when exhibitions feeds were dealt with differently to the other feeds. Now that we download media for these two we can remove it.

app/models/feed.rb
lbiedinger: nope this should be removed

app/views/application_view.rb
lbiedinger: this was to stop facebook errors in JS when running tests. I thought I had already removed it from this PR. Will do so again.

lbiedinger: in fact I think since https://github.com/europeana/Europeana-Patternlab/pull/245 has gone in we may be less likely to have facebook .js errors during testing at all.

...maybe.

PR #1008, opened 2017-10-19
Scheduled Gallery validation

lbiedinger:
Re: https://europeana.atlassian.net/browse/EC-2236

Adds a url column to the gallery_image table, this will store the url for the image hosted by the provider.

Adds a scheduled job to verify all the urls for all the published galleries are retrievable and when they are retrieved return images.


db/migrate/20171018122656_add_edm_is_shown_by_to_gallery_images.rb
```diff
@@ -0,0 +1,31 @@
...
+      save
+    rescue
+      delete
+    end
+  end
+
+  def up
+    add_column :gallery_images, :edm_is_shown_by, :text
+
+    GalleryImage.all.each(&:set_edm_is_shown_by)
```
rwd: The performance of this migration can be improved by a factor of 100 by querying the API for 100 records at a time:
```ruby
GalleryImage.find_in_batches(batch_size: 100) do |batch|
  record_ids = batch.map(&:europeana_record_id)
  api_query = Europeana::Record.search_api_query_for_record_ids(record_ids)
  response = Europeana::API.record.search(query: api_query, profile: 'rich', rows: 100)

  batch.each do |image|  
    item = response['items'].detect { |item| item['id'] == image.europeana_record_id }
    if item.nil?
      image.delete
    else
      image.edm_is_shown_by = item['edmIsShownBy'].first
      image.save
    end
  end
end
```

db/schema.rb
```diff
@@ -206,6 +206,7 @@
     t.integer  "position"
     t.datetime "created_at",          null: false
     t.datetime "updated_at",          null: false
+    t.text     "edm_is_shown_by",     null: false
```
rwd: Given that we plan to introduce the feature to also be able to specify a `hasView` image from a record, should we instead name this column `url`?

app/views/gallery_validation_mailer/post.text.erb
```diff
@@ -0,0 +1,9 @@
+Automated Gallery Validation for Europeana Gallery: <%= gallery_url(:en, @gallery) %>
+<% @validation_errors.each_pair do |key, errors| %>
+  Validation error(s) for the image '<%= document_url(:en, key.sub('/','')) %>':
```
rwd: Add a space between the args to `sub`

app/mailers/gallery_validation_mailer.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class GalleryValidationMailer < ApplicationMailer
+  def post(gallery:, validation_errors:)
+    @gallery = gallery
```
rwd: `fail` here if `Rails.application.config.x.gallery_validation_mail_to` is not set.

app/jobs/gallery_validation_job.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+##
+# Validate that a gallery has only properly displaying images
+# Should an issue be discovered, a slack channel will be notified by sending an email.
+class GalleryValidationJob < ApplicationJob
+  queue_as :default
+
+  def perform(gallery_id)
+    @gallery = Gallery.find(gallery_id)
```
rwd: `fail` here if `Rails.application.config.x.gallery_validation_mail_to` is not set.

app/jobs/gallery_validation_job.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+##
+# Validate that a gallery has only properly displaying images
+# Should an issue be discovered, a slack channel will be notified by sending an email.
+class GalleryValidationJob < ApplicationJob
```
rwd: I don't see any code actually queueing this job...

app/jobs/gallery_validation_job.rb
```diff
@@ -0,0 +1,64 @@
...
+    errors = []
+    europeana_id = gallery_image.europeana_record_id
+    stored_image_url = gallery_image.image_url
+    api_document = api_search_response.detect { |record| record['id'] == europeana_id }
+    if api_document
+      unless api_document['edmIsShownBy'] == stored_image_url
+        errors << "edm:isShownBy for image '#{europeana_id}' has changed"
+      end
+      unless retrievable_image?(stored_image_url)
+        errors << "The image '#{europeana_id}' can not be retrieved at '#{stored_image_url}'."
```
rwd: This should differentiate between non-retrievable URLs (404, 500, etc) and retrievable URLs which are not images.

app/models/gallery.rb
```diff
@@ -75,12 +76,15 @@ def to_param
   def set_images_from_portal_urls
     transaction do
       enumerable_image_portal_urls.map { |url| Europeana::Record.id_from_portal_url(url) }.tap do |record_ids|
+        matched_ids = []
         record_ids.each_with_index do |record_id, i|
-          GalleryImage.find_or_create_by(gallery: self, europeana_record_id: record_id).tap do |image|
+          image_url = response_items.detect { |record| record['id'] == record_id }['edmIsShownBy'].first
+          GalleryImage.find_or_create_by!(gallery: self, europeana_record_id: record_id, image_url: image_url).tap do |image|
```
rwd: Should this really be `GalleryImage.find_or_create_by!` instead of `GalleryImage.find_or_create_by`? Won't the former prevent display of errors for all invalid image URLs at once?

app/views/application_view.rb
```diff
@@ -59,7 +59,7 @@ def head_links
   end
 
   def fb_campaigns_on
-    true
+    true unless Rails.env.test?
```
rwd: What's this?

db/migrate/20171018122656_add_image_url_to_gallery_images.rb
```diff
@@ -0,0 +1,32 @@
...
+    add_column :gallery_images, :image_url, :text
+
+    GalleryImage.find_in_batches(batch_size: 100) do |batch|
+      record_ids = batch.map(&:europeana_record_id)
+      api_query = Europeana::Record.search_api_query_for_record_ids(record_ids)
+      response = Europeana::API.record.search(query: api_query, profile: 'rich', rows: 100)
+
+      batch.each do |image|
+        response_item = response['items'].detect { |item| item['id'] == image.europeana_record_id }
+        image.image_url = response_item.nil? ? 'UNKNOWN' : response_item['edmIsShownBy'].first
```
rwd: We need to check for the presence of `response_item['edmIsShownBy']` before calling `response_item['edmIsShownBy'].first` as not all records have `edm:isShownBy` and this would then fail.

app/jobs/gallery_validation_job.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  private
+
+  def validate_gallery_image(gallery_image)
+    errors = []
+    europeana_id = gallery_image.europeana_record_id
+    stored_image_url = gallery_image.image_url
+    api_document = api_search_response.detect { |record| record['id'] == europeana_id }
+    if api_document
+      unless api_document['edmIsShownBy'] == stored_image_url
```
rwd: `api_document['edmIsShownBy']` will either be `nil` or an `Array`, not a `String`.

app/models/gallery.rb
```diff
@@ -75,12 +76,15 @@ def to_param
   def set_images_from_portal_urls
     transaction do
       enumerable_image_portal_urls.map { |url| Europeana::Record.id_from_portal_url(url) }.tap do |record_ids|
+        matched_ids = []
         record_ids.each_with_index do |record_id, i|
-          GalleryImage.find_or_create_by(gallery: self, europeana_record_id: record_id).tap do |image|
+          image_url = response_items.detect { |record| record['id'] == record_id }['edmIsShownBy'].first
```
rwd: `edmIsShownBy` may be nil.

app/jobs/gallery_validation_job.rb
lbiedinger: Does this really matter to an editor?

app/jobs/gallery_validation_job.rb
```diff
@@ -0,0 +1,64 @@
...
+      api_query = Europeana::Record.search_api_query_for_record_ids(@gallery.images.map(&:europeana_record_id))
+      Europeana::API.record.search(query: api_query, profile: 'rich', rows: 100)['items'] || []
+    end
+  end
+
+  def retrievable_image?(is_shown_by)
+    response = download_response(is_shown_by)
+    return false unless response
+    content_type = response.headers[:content_type]
+    return false unless content_type&.start_with?('image')
```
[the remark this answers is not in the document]
lbiedinger: yes it's shorthand for a  nil check

app/models/gallery.rb
lbiedinger: not if the validation ran successfully, which happens before save, this whole method happens after save.
we could also set it to 'UNKNOWN' if we think it's going to be called outside of the normal cms or by some other script?

app/views/application_view.rb
lbiedinger: This was to stop facebook errors in test. we should do this on develop probably, so I will remove it from this PR as it's not specifically related.

db/migrate/20171018122656_add_image_url_to_gallery_images.rb
```diff
@@ -0,0 +1,32 @@
...
+class AddImageUrlToGalleryImages < ActiveRecord::Migration
+  class GalleryImage < ActiveRecord::Base
+    validates :gallery_id, presence: true
+    validates :europeana_record_id,
+              presence: true, format: { with: Europeana::Record::ID_PATTERN }
+    validates :image_url, presence: true
+  end
+
+  def up
+    add_column :gallery_images, :image_url, :text
```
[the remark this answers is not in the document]
lbiedinger: I wanted to make explicit that this is the URL for the provider hosted image, but sure.
Be careful with your migrations if you've ran this locally already since I renamed it.

app/jobs/gallery_validation_job.rb
lbiedinger: added a call to clock rb and a job so we can run it manually to jobs.rake
https://github.com/europeana/europeana-portal-collections/pull/1008/commits/88821aa2a0b3a599f60bef62aa6bd1c108259a0a#diff-da3f4822b15eb7097ba27f136e6c8f3c

app/models/gallery.rb
rwd: Right you are. That's fine as-is then.

.env.example
```diff
@@ -157,6 +160,10 @@
 ## Time each day at which to run the facet link group harvesting job. No default.
 # SCHEDULE_FACET_ENTRY_GROUPS_GENERATOR="05:30"
 
+## Times each day at which to run gallery validation, multiple values are
+# separated by spaces
+#SCHEDULE_GALLERY_VALIDATION="08:00 15:00"
```
rwd: As well as the schedule and Rake task, do you think we should trigger the validation job immediately after each gallery is created or published? To run some of the more extensive checks like content type and download verification right away.

lib/clock.rb
```diff
@@ -37,6 +39,12 @@
     end
   end
 
+  every(1.day, 'gallery-validation', at: ENV['SCHEDULE_GALLERY_VALIDATION']) do
+    Gallery.published.each do |gallery|
+      GalleryValidationJobJob.perform_later(gallery.id)
```
rwd: `GalleryValidationJob` not `GalleryValidationJobJob`

lib/tasks/jobs.rake
```diff
@@ -49,4 +49,11 @@ namespace :jobs do
       FacetLinkGroupGeneratorJob.perform_later facet_link_group
     end
   end
+
+  desc 'Queue GalleryValidationJob'
+  task gallery_validation: :enironment do
+    Gallery.published.each do |gallery|
+      GalleryValidationJobJob.perform_later(gallery.id)
```
rwd: `GalleryValidationJob` not `GalleryValidationJobJob`

spec/models/gallery_spec.rb
```diff
@@ -1,55 +1,7 @@
 # frozen_string_literal: true
 require 'models/concerns/is_permissionable_examples'
 RSpec.describe Gallery do
```
rwd: We should have tests to ensure that `Gallery#set_images_from_portal_urls` results in `GalleryImage` records with `url` attribute set from `edmIsShownBy`.

.env.example
lbiedinger: yes, I checked this with Adrian and he thinks it would be useful

lib/tasks/jobs.rake
lbiedinger: good jobjob you spotted this

spec/models/gallery_spec.rb
lbiedinger: tests added here: 7d0d1f27c9ba19c3992065c2a02d23e866bc4e91

PR #1010, opened 2017-10-20
Re: #2472 show new item page when param 'new_item_page' is true

andyjmaclean:


app/presenters/document/field_group_presenter.rb
```diff
@@ -33,7 +33,8 @@ def display
 
       {
         title: display_title,
-        sections: display_sections
+        sections: display_sections,
+        section_id: @id
```
rwd: Having two properties `sections` and `section_id` is confusing. Shouldn't the latter be named `group_id` or even just `id`?

rwd:
I have refactored this approach in https://github.com/europeana/europeana-portal-collections/pull/1011

PR #1011, opened 2017-10-23
New item page design integration

rwd:
Refactoring https://github.com/europeana/europeana-portal-collections/pull/1010


andyjmaclean:
@rwd I've renamed "section_id" to "id" as per your recommendation.

PR #1013, opened 2017-10-25
Gallery annotations

rwd:
Re: https://europeana.atlassian.net/browse/EC-1957


app/jobs/store_gallery_annotations_job.rb
```diff
@@ -0,0 +1,84 @@
...
+  def create_annotations
+    gallery.images.each do |image|
+      unless api_annotation_targets.include?(image.annotation_target)
+        logger.info("Creating annotation linking #{image.annotation_target} to #{gallery.annotation_link_resource_uri}".green.bold)
+        Europeana::API.annotation.create(create_annotation_api_params(image))
+      end
+    end
+  end
+
+  def create_annotation_api_params(image)
```
lbiedinger: if we move this into the `AnnotationsApiConsumer` alongside: 
`annotations_api_delete_params`
and rename it to `annotations_api_create_params`
this would expose more consistent CRUD interaction through the `AnnotationsApiConsumer` concern.
Then it would make sense though not to rely on `image.annotation_body` but rather have annotations_api_create_params take only a `body` arguement for now.

app/controllers/concerns/europeana/annotations_api_consumer.rb
```diff
@@ -20,11 +19,22 @@ def document_annotations(document_id)
...
     end
 
-    def annotations_search_for_document(document_id)
-      Europeana::API.annotation.search(annotations_api_search_params(document_id))
+    def annotations_for_record(record_id, **local_params)
+      search_response = annotations_search_for_document(record_id, local_params)
+      annotations_from_search_response(search_response)
+    end
+
+    def annotations_search_for_document(record_id, **local_params)
```
lbiedinger: should this be `annotations_search_for_record` ?

app/controllers/concerns/europeana/annotations_api_consumer.rb
```diff
@@ -20,11 +19,22 @@ def document_annotations(document_id)
...
+      local_params[:qf] ||= []
+      local_params[:qf] += [
+        %(generator_name:#{Rails.application.config.x.europeana[:annotations].api_generator_name || 'Europeana.eu*'}),
+        %(target_uri:"http://data.europeana.eu/item#{record_id}")
+      ]
+      Europeana::API.annotation.search(annotations_api_search_params(record_id, local_params))
```
lbiedinger: this isn't working on the record page, it shouldn't pass record_id along

app/models/europeana/record/annotations.rb
```diff
@@ -0,0 +1,32 @@
...
+    module Annotations
+      # TODO: handle pagination if more than 100 items
+      def annotations
+        @annotations ||= Annotation.find(annotations_search_params)
+      end
+
+      def annotations_search_params
+        {
+          qf: [
+            %(generator_name:#{annotations_api_generator_name}),
```
lbiedinger: Not sure if it makes sense in the context of this PR, or if it's limited by the Annotation API and how that stores the different kind of annotations, but it would be nice if this was more consistent and we used creator_name here too.

EuropeanaDevelopment: `creator_name` is derived from the user token, `generator_name` from the API key.

For gallery annotations we only want to retrieve those created by the Collections gallery annotator, but on item pages we want annotations from potentially many users.

i.e. this needs to remain as it is.

app/models/gallery/annotations.rb
```diff
@@ -0,0 +1,36 @@
...
+        qf: [
+          'creator_name:"Europeana.eu Gallery"',
+          'link_relation:isGatheredInto',
+          'motivation:linking',
+          %(link_resource_uri:"#{annotation_link_resource_uri}")
+        ]
+      }
+    end
+
+    # TODO: do not hard-code, but use deployment's URL, without locale? or make
```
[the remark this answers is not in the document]
EuropeanaDevelopment: If present, yes. If not, default to www.europeana.eu.

app/models/europeana/record/annotations.rb
rwd: `creator_name` is derived from the user token, `generator_name` from the API key.

For gallery annotations we only want to retrieve those created by the Collections gallery annotator, but on item pages we want annotations from potentially many users.

i.e. this needs to remain as it is.

app/models/gallery/annotations.rb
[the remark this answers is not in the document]
rwd: If present, yes. If not, default to www.europeana.eu.

PR #1014, opened 2017-10-31
Use https:// for default URL to exhibitions host

rwd:


rwd:
Will create a Jira issue for this and extend to address some related problems.

PR #1018, opened 2017-11-01
Set googleAnalyticsLinkedDomains JS var from env

rwd:
Re: https://europeana.atlassian.net/browse/EC-2452

Env var is `GOOGLE_ANALYTICS_LINKED_DOMAINS`


spec/views/collections/show_spec.rb
```diff
@@ -20,34 +19,40 @@
   subject { rendered }
 
   it 'should have meta description' do
+    render
```
lbiedinger: rather than adding `render` to each example, let's wrap all the examples in:
```
  describe 'the rendered page' do
    before do
      render
    end
   ...
  end
```

This will be DRYer and still passes all the tests.

rwd: Given that having `render` in a before block elsewhere caused issues, I'm inclined to leave `render` called in each test.

lbiedinger: fair enough

PR #1022, opened 2017-11-08
Add record gallery.json method

rwd:
Re: https://europeana.atlassian.net/browse/EC-2503

Also:
* Use thumbnail API for gallery thumbs, not S3
* Stop querying Search API on gallery foyer, use stored image URLs instead

### DIFF
--- a/app/controllers/galleries_controller.rb
+++ b/app/controllers/galleries_controller.rb
@@ -13,7 +13,6 @@ class GalleriesController < ApplicationController
     @galleries = Gallery.includes(:images).published.order(published_at: :desc).
                  page(pagination_page).per(pagination_per).with_topic(gallery_topic)
     @selected_topic = gallery_topic
-    images = gallery_images_for_foyer(@galleries)
     @hero_image = homepage_hero_image
 
     @body_cache_key = foyer_body_cache_key(topic: @selected_topic, per: @galleries.limit_value, page: @galleries.current_page)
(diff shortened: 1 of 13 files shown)


app/controllers/portal_controller.rb
```diff
@@ -112,6 +112,16 @@ def annotations
     end
   end
 
+  def gallery
+    @gallery = Gallery.published.joins(:images).
```
lbiedinger: I don't think this is really the way https://europeana.atlassian.net/browse/EC-1901 intends the galleries to be included on item pages. I thought we were going to load in any galleries that are included in annotations?
So this method should be for 'galleries' not 'gallery' and the retrieve all the galleries via the annotations.

app/controllers/galleries_controller.rb
```diff
@@ -16,7 +17,6 @@ def index
     @hero_image = homepage_hero_image
 
     @body_cache_key = foyer_body_cache_key(topic: @selected_topic, per: @galleries.limit_value, page: @galleries.current_page)
-    @documents = search_api_for_image_metadata(images) unless body_cached?
```
lbiedinger: Removing this line and constructing the gallery tiles/images structure in the view via the new helper means we don't need to assign images on line :16 also that will remove the codeclimate issue on this PR.
plus we can remove the 'gallery_images_for_foyer' method from the controller.

PR #1026, opened 2017-11-15
add collections rss feed

lbiedinger:


app/views/collections/index.rss.builder
```diff
@@ -0,0 +1,24 @@
...
+    xml.lastBuildDate(DateTime.now.rfc2822)
+    xml.tag!('atom:link', rel: 'self', type: 'application/rss+xml', href: collections_url(format: 'rss'))
+
+    displayable_collections.each do |collection|
+      xml.item do
+        xml.title(collection.title)
+        xml.link(collection_url(collection))
+        xml.pubDate(collection.updated_at.rfc2822)
+        xml.guid(collection_url(collection))
+        xml.description(collection.landing_page.strapline)
```
rwd: This needs to populate the `%{total_item_count}` placeholder.

app/views/collections/index.rss.builder
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+# Helper method for the collections description
+# derived from the associated landing page strapline.
+def strapline(collection)
```
rwd: Let's move this into `CollectionsHelper` as `collection_strapline`, then make both this RSS view and the `Collections::Show` Mustache view class use it, to keep it DRY, removing the assignment of `@total_item_count` from `CollectionsController#show`.

spec/controllers/collections_controller_spec.rb
```diff
@@ -1,15 +1,25 @@
...
+      end
+
+      it 'redirects to home' do
+        expect(response).to redirect_to(home_url)
+      end
     end
 
-    it 'redirects to home' do
-      expect(response).to redirect_to(home_url)
+    context 'when requesting as an rss feed' do
```
rwd: Could we spec that only published collections are included?

lbiedinger: Added feature spec for the  collections rss feed that includes these checks

PR #1027, opened 2017-11-20
Use Jbuilder template for JSON search results

rwd:
Plus:
* Removed some redundant Blacklight-related methods from `CollectionsController`
* Moved search interaction logging logic into `SearchInteractionLogging` controller concern
* Do not populate collection-related data when it won't be used by the view
* Move error message construction for `FederationsController#show` into its Jbuilder template, and stop lazy use of `rescue` without an exception class specified


app/controllers/collections_controller.rb
```diff
@@ -18,35 +19,24 @@ def index
...
       (@response, @document_list) = search_results(params)
-      log_search_interaction(
-        search: params.slice(:q, :qe, :qf, :f, :mlt, :range).inspect,
-        total: @response.total
-      )
+      log_search_interaction_on_search(@response)
+    else
+      @collection_stats = collection_stats
+      @recent_additions = recent_additions
+      @total_item_count = Rails.cache.fetch("record/counts/collections/#{@collection.key}")
```
lbiedinger: this could use the cached_record_count method from the RecordCountsHelper

PR #1029, opened 2017-11-21
Fetch dataProvider direct from document

rwd:
To prevent double HTML escaping in combination with Blacklight.

Re: https://europeana.atlassian.net/browse/EC-2524

### DIFF
--- a/app/presenters/concerns/blacklight_document_presenter.rb
+++ b/app/presenters/concerns/blacklight_document_presenter.rb
@@ -25,12 +25,10 @@ module BlacklightDocumentPresenter
   end
 
   def field_value(fields, **options)
-    unescape = options[:unescape]
     method = options[:context] == :index ? :render_index_field_value : :render_document_show_field_value
 
     [fields].flatten.each do |field|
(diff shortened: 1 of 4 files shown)


app/presenters/concerns/blacklight_document_presenter.rb
```diff
@@ -8,18 +8,20 @@ module BlacklightDocumentPresenter
...
+    def render_values(values, field_config = nil)
+      options = {}
+      options = field_config.separator_options if field_config && field_config.separator_options
 
-    values.to_sentence(options)
+      values.to_sentence(options)
+    end
   end
 
   def field_value(fields, **options)
```
lbiedinger: I think we can get rid of the `unescape` variable and the call to `CGI.unescapeHTML` in here altogether then since it's never passed as an option now.

lbiedinger: at least I couldn't find anywhere that we would pass it as an option. So unless you envisage explicitly calling this in the future, let's remove it from this method.

PR #1036, opened 2017-12-06
Revert to cookie-based session storage

rwd:
Re: https://europeana.atlassian.net/browse/EC-2581


rwd:
This needs to be based on master, not develop.

PR #1038, opened 2017-12-06
Revert to cookie session storage

rwd:
Re: https://europeana.atlassian.net/browse/EC-2581


db/schema.rb
```diff
@@ -209,7 +206,6 @@
     t.integer  "position"
     t.datetime "created_at",          null: false
     t.datetime "updated_at",          null: false
-    t.text     "url",                 null: false
```
lbiedinger: this is breaking tests and galleries. I don't see why it's not in the schema either.

rwd: Yep, not sure why that got removed. Fixed in c85e8103a827c6827b8d7cc08d6c185a4c3847ef

PR #1045, opened 2018-01-11
EC-2596 New media fields / bump translations / bump styleguide

lbiedinger:


app/presenters/concerns/document/metadata/rights.rb
```diff
@@ -31,8 +31,9 @@ def simple_rights(rights)
           }
         else
           license_flag_key = rights.template_license.present? ? rights.template_license : rights.id.to_s.upcase
-
           {
+            license_brief: rights.reusability ? t('global.facet.reusability.' + rights.reusability + '-brief') : false,
```
rwd: This and the following call to `#t` should use the `scope` option, as with the call for `license_human`.

rwd:
@lbiedinger NM, I see you have a PR to do so already!

PR #1052, opened 2018-02-15
Fix edm rights from api for (vocab|page) regex.

lbiedinger:


lbiedinger:
Duplicate functionality of more extensive version in: https://github.com/europeana/europeana-portal-collections/tree/hotfix/EC-2684-edm-rights-for-api-query

PR #1059, opened 2018-03-22
EC-2695 Leave parentheses in quoted value presenters.

lbiedinger:


spec/presenters/document/field_group_presenter_spec.rb
```diff
@@ -181,6 +181,20 @@
           it 'links to a search for the quoted field value' do
             expect(CGI.unescape(subject[:sections].first[:items].first[:url])).to eq('/en/search?q=what:"Photography"')
           end
+
+          context 'with parentheses in the defenition' do
```
rwd: `s/defenition/value/`

PR #1062, opened 2018-03-27
Support non-AWS S3 storage

rwd:


rwd:
To be recreated without breaking `bundle update`.

PR #1063, opened 2018-03-28
Non-AWS S3 & stack update

rwd:


rwd:
Recreated as #1064 without `rubocop -a`.

PR #1067, opened 2018-03-28
Feat/ec 2755 datacentre failure notice

andyjmaclean:


rwd:
This branch needs to be based on master and the PR made against it, to hotfix.

PR #1078, opened 2018-05-02
Gallery images by web resource

rwd:
Summary:
This introduces functionality so that each image in a gallery may be any of the edm:isShownBy or edm:hasView web resources of a Europeana record (instead of always using the edm:isShownBy).

Changes:
* `Gallery` model:
  * Adds two new stored fields: `image_portal_urls` as an array of the URLs extracted from those input by an editor in RailsAdmin; `image_errors` as a JSONB representation of a hash of errors detected during asynchronous validation of the portal URLs
  * When inputting image portal URLs via RailsAdmin, the editor specifies an edm:hasView by appending `?view=http://example.com/hasView.jpg` (inserting the URI of the specific web resource); any special URL characters like `&` and `?` in this parameter will need to be escaped; if no `view` parameter is specified, the edm:isShownBy of the record will be used
  * When saved, galleries will no longer attempt to run any validation for image portal URLs against the Europeana Record API.
  * When saved, galleries will no longer create/update/delete associated `GalleryImage`s
  * Prevented from being published without the required number of verified images
* `GalleryImage` model:
  * Introduces three modules for: annotations; validation of URLs via the Europeana Record API; validation of URLs via HTTP requests (for the web resources themselves)
  * HTTP validation of web resources is via the [media proxy](https://github.com/europeana/europeana-proxy-ruby)
* `GalleryDisplayabilityJob` (renamed from `GalleryValidationJob`):
  * triggers full verification of each image in a gallery including against the Record API and web resource HTTP URLs
  * Propagates verified images to the gallery's associated `GalleryImage`s
  * Removes images failing verification, without removing their portal URL from the gallery, and records errors in `Gallery#image_errors` for display to the editor in the CMS
  * Renamed Rake task `jobs:gallery_validation` to `jobs:galleries:displayability`
* [Annotations](https://pro.europeana.eu/resources/apis/annotations):
  * Creates annotations with the target being a `SpecificResource` where the `scope` is the URI of the record and the `source` is the URI of the web resource
  * Adds support for setting the portal hostname to use in annotations created for gallery images  an environment variable `GALLERY_ANNOTATION_LINK_RESOURCE_HOST`
  * When searching for annotations for a record, use `target_record_id` to also detect `SpecificResource` annotations
  * Rake task `jobs:galleries:annotations:create` to generate annotations for all published galleries
  * Rake task `jobs:galleries:annotations:delete` to delete annotations for all published galleries

Dependencies:
* Requires PostgreSQL version 9.4+

Deployment:
* Run `bundle exec rake db:migrate`


app/jobs/gallery_validation_job.rb
```diff
@@ -7,59 +7,37 @@ class GalleryValidationJob < ApplicationJob
...
-    stored_image_url = gallery_image.url
-    api_document = api_search_response.detect { |record| record['id'] == europeana_id }
-    if api_document
-      unless api_document['edmIsShownBy']&.first == stored_image_url
-        errors << 'edm:isShownBy has changed'
-      end
-      unless retrievable_image?(stored_image_url)
-        errors << "can not retrieve #{stored_image_url}"
-      end
+    if @validation_errors.present?
```
lbiedinger: validate_number_of_image_portal_urls

spec/models/gallery_spec.rb
```diff
@@ -81,160 +95,204 @@
...
+      subject { described_class.new(image_errors: { 'url1' => ['err1'], 'url2' => ['err2'] }) }
+
+      it 'is propagated to errors on image_portal_urls_text after initialize' do
+        expect(subject.errors[:image_portal_urls_text]).to include('err1')
+        expect(subject.errors[:image_portal_urls_text]).to include('err2')
+      end
+    end
+  end
+
+#   describe 'per-image API response validation' do
```
lbiedinger: these are the old specs right? If so they can be removed.

PR #1091, opened 2018-06-08
Optimise facet URL generation

rwd:
Using `String` because using `Blacklight::SearchState` and `Hash` is too expensive for large numbers of facet values.

### DIFF
--- a/app/presenters/facet/collection_presenter.rb
+++ b/app/presenters/facet/collection_presenter.rb
@@ -4,21 +4,14 @@ module Facet
   class CollectionPresenter < SimplePresenter
     def add_facet_url(item)
       value = facet_value_for_facet_item(item)
-
       base_url = default_facet_value?(value) ? search_url : collection_url(value)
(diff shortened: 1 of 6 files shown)


app/presenters/facet/collection_presenter.rb
```diff
@@ -2,26 +2,16 @@
...
-    # as it redirects to the standard search.
-    # The only reason it takes a facet_item as a param is because
-    # this overrides {FacetPresenter}
-    #
-    # @param see {#facet_item}
-    # @return [Hash] Request parameters without the collection
-    def remove_facet_params(_item)
-      params.except(:id).merge(controller: :portal, action: :index)
+    # TODO: is this useful behaviour when the collection facet is presented
+    #       like a radio button where one is always selected?
```
lbiedinger: IMO yes it's useful, because it then behaves like the other facets. Where clicking on a selected facet will deselect that facet.

the 'ordered' method down below, however is not being used as far as I can tell.

PR #1100, opened 2018-06-28
Record sets browse pages

rwd:
The primary purpose of this PR is to introduce a new type of page which editors may create in the CMS: the "record sets browse page". These pages consist of one or more sets of Europeana records, curated by the editor, and presented using the new grid layout. When creating these pages, the editor need only provide the portal URL for each record, and metadata required for the page's presentation will be retrieved on-demand from the Search API (and then the rendered HTML cached).

This PR also undertakes some refactoring to:
1. Reduce our reliance on Blacklight for retrieval and presentation of pages of search results, resulting in the new page type not using Blacklight, and the galleries also no longer using Blacklight.
2. Start replacing the custom `HasSettingsAttribute` model concern with `ActiveRecord::Store`

Deployment:
* `bundle exec rake db:migrate`
* `bundle exec rake db:seed`


db/migrate/20180628102730_add_link_text_setting_to_page_translations.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class AddLinkTextSettingToPageTranslations < ActiveRecord::Migration
+  def change
+    reversible do |dir|
+      dir.up do
+        Page::Browse::RecordSets.add_translation_fields! settings_link_text: :string
```
lbiedinger: I think, this breaks saving of any other pages(Page, Page::Landing, Page::Error) through rails admin.
Are you sure a translation can be added to a setting?

app/models/europeana/record/set.rb
```diff
@@ -0,0 +1,106 @@
...
+      # @param value [Array<String>]
+      def portal_urls=(value)
+        @portal_urls_to_europeana_ids = value.each_with_object({}) do |url, memo|
+          memo[url] = Europeana::Record.id_from_portal_url(url)
+        end
+        self.europeana_ids = @portal_urls_to_europeana_ids.values.compact
+      end
+
+      def portal_urls_text
+        @portal_urls_text || portal_urls&.join("\n\n")
```
lbiedinger: when adding a new set this adds a single white space to the text area. It's actually kind of annoying when pasting in urls since they don't validate if theres a leading space. It wouldn't be a problem either if before validating URLs the value has leading and trailing spaces removed down in portal_urls_text=(vlaue)

rwd: I do not understand what you mean here and/or can not replicate this problematic behaviour.

app/controllers/concerns/europeana/search_api_consumer.rb
```diff
@@ -0,0 +1,44 @@
...
+      id_count = europeana_ids.count
+      return {} if id_count.zero?
+
+      per_page = 100
+      pages = id_count / per_page + 1
+      paged_ids = europeana_ids.each_slice(per_page).to_a
+
+      {}.tap do |results|
+        page = 0
+        while page < pages
```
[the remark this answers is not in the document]
rwd: Good point about the range, but I'll use an `#each` iterator instead of the for loop.

spec/controllers/pages_controller_spec.rb
```diff
@@ -36,5 +36,19 @@
...
+    context 'when type is Page::Browse::RecordSets' do
+      let(:page) { pages(:newspapers_a_to_z_browse) }
+      let(:params) { { locale: 'en', format: 'html', page: page.slug } }
+
+      it 'renders browse page template' do
+        expect(response.status).to eq(200)
+        expect(response).to render_template('pages/browse/record_sets')
+      end
+
+      it 'queries API for documents'
```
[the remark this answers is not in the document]
rwd: Nope. Rectified.

app/models/europeana/record/set.rb
lbiedinger: I was referring to the behaviour in rails-admin. I don't think it has anything to do with this line actually after trying to change it here.

To reproduce:
1) Go to the edit page of a browse page.
2) Click the add section button.
3) Observe there's a single empty space in the 'Portal URLs' text area.
4) Add a portal URL to the text area(without removing the space)
5) Click Save
6) Observe that it fails, since the first URL is invalid.

Perhaps this behaviour can be circumvented in the RailsAdmin initializer. I'm not sure how though since it only affects blank values. So that's why trimming the input may be the better solution.

app/models/europeana/record/set.rb
```diff
@@ -25,7 +23,7 @@ class Set < ActiveRecord::Base
       validates :title, presence: true
       validate :validate_portal_urls_presence, :validate_portal_urls_format
 
-      has_settings :query_term
+      store_accessor :settings, :query_term
```
lbiedinger: should the key here be ':config' the same as for the other models?

rwd: No. We will keep using `:settings` but to switch `Page` to using ActiveRecord::Store in a non-disruptive manner need to name its new serialized settings store something else first, i.e. `:config`, then later rename that back to `:settings`.

PR #1105, opened 2018-06-29
Creator info for new record page

lbiedinger:


app/presenters/document/record_presenter.rb
```diff
@@ -33,6 +33,39 @@ def creator_title
       end
     end
 
+    ##
+    # Looks up dc:Creators, if they are europeana entity agents returns them paired with
+    # the respective urls for retrieving thumbnails via JS.
+    #
+    # @return [Hash]
```
rwd: Or an array of hashes, i.e. `@return [Hash,Array<Hash>]`

app/presenters/document/record_presenter.rb
```diff
@@ -33,6 +33,39 @@ def creator_title
...
+
+      if !creator_entities&.any?
+        { title: field_value(['proxies.dcCreator']), europeana_entities: false }
+      else
+        creator_entities.map do |agent|
+          {
+            title: document.localize_lang_map(agent['prefLabel']),
+            url: agent['about']
+          }
+        end << { europeana_entities: true }
```
rwd: I think this will be awkward for the front-end to detect, as the final hash in this array of hashes.

app/presenters/document/record_presenter.rb
```diff
@@ -33,6 +33,39 @@ def creator_title
       end
     end
 
+    ##
+    # Looks up dc:Creators, if they are europeana entity agents returns them paired with
+    # the respective urls for retrieving thumbnails via JS.
+    #
+    # @return [Hash]
+    def creators_info
+      creator_entities = agents_for('proxies.dcCreator').select! do |entity|
```
rwd: Just `#select` will suffice here.

app/presenters/document/record_presenter.rb
```diff
@@ -33,6 +33,39 @@ def creator_title
...
+        URI.parse(entity['about']).host == 'data.europeana.eu'
+      end
+
+      if !creator_entities&.any?
```
rwd: `if creator_entities.blank?` is clearer

app/presenters/document/record_presenter.rb
```diff
@@ -33,6 +33,39 @@ def creator_title
...
+        { title: field_value(['proxies.dcCreator']), europeana_entities: false }
```
rwd: The argument to `field_value` does not need to be an array.

app/presenters/document/record_presenter.rb
```diff
@@ -33,6 +33,39 @@ def creator_title
...
+        end << { europeana_entities: true }
+      end
+    end
+
+    ##
+    # Looks up an agent entity for the field passed as a param.
+    #
+    # @param field [#string] The field in the document to find an agent for
+    # @return [Array<Hash>]
+    def agents_for(field)
```
rwd: Should we be reusing/extending the existing entity handling in `Document::EntityPresenter` and `Document::FieldGroup::SectionPresenter`?

app/presenters/document/record_presenter.rb
```diff
@@ -33,6 +33,39 @@ def creator_title
...
     end
 
+    ##
+    # Looks up dc:Creators, if they are europeana entity agents returns them paired with
+    # the respective urls for retrieving thumbnails via JS.
+    #
+    # @return [Hash]
+    def creators_info
+      creator_entities = agents_for('proxies.dcCreator').select do |entity|
+        URI.parse(entity['about']).host == 'data.europeana.eu'
```
rwd: We will need to rescue exceptions from `URI.parse` as agent URIs may be in a format it does not understand.

app/presenters/document/record_presenter.rb
rwd: ... which may also be awkward for the front-end to handle both...

app/presenters/document/record_presenter.rb
rwd: ... or just use `entity['about'].include?('://data.europeana.eu/')

PR #1108, opened 2018-07-09
Creators as field group

lbiedinger:
These changes modify the value_presenter logic in order to facilitate using a new field group in order to display the 'creators' section on the new-item page.


app/presenters/document/record_presenter.rb
```diff
@@ -50,13 +51,15 @@ def creators_info
```
rwd: I worry that there is too much duplication in functionality / overlap in responsibility between this method and `Document::FieldGroup::Section::ValuePresenter#display`, which for `field_group('creators')` on record `/92034/GVNRC_NMI01_13925` would return:
```ruby
{
  :title => "translation missing: en.site.object.meta-label.creators",
  :sections => [
    {
      :title => "Creator",
      :items => [
        {
          :text => "Alphons Diepenbrock",
          :url => "/portal/en/explore/people/157024-alphons-diepenbrock.html",
          :ga_data => nil,
          :extra_info => nil,
          :raw => false
        }
      ],
      :is_desc => false
    }
  ],
  :id => "creators"
}
```
... which already contains most of what `#creators_info` generates.

I would suggest adding to the return value of `Document::FieldGroup::Section::ValuePresenter#display`:
```ruby
# ...
json_url: json_url,
entity: for_entity?,
europeana_entity: for_europeana_entity?
# ...
```
... defining the new methods accordingly.

The template would then of course need to be adapted to handle this structure instead.

If possible, it would be prefeable to eliminate the need for this `creators_info` method, as we had many field group specific methods in the past, but refactored into the YAML config file for field groups with reusable display logic.

app/presenters/concerns/document/entities.rb
```diff
@@ -24,5 +24,9 @@ def entities_for_type(type, field = nil)
       end
       typed_entities || []
     end
+
+    def entity_labels_with_field_value_fallback(field)
```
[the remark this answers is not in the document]
lbiedinger: nothing anymore.

app/helpers/entities_helper.rb
```diff
@@ -41,4 +41,14 @@ def entity_url_slug(entity)
...
   end
+
+  def portal_entity_path(entity_url, slug: nil, format: 'html')
+    return nil unless europeana_entity_url?(entity_url)
+    entity_uri_parts = URI.parse(entity_url).path.split('/')[1..-1]
+    controller.entity_path(type: entities_human_type(entity_uri_parts[0]), id: entity_uri_parts.last, slug: slug, format: format)
+  end
+
+  def europeana_entity_url?(entity_url)
+    %r(https?://data.europeana.eu/[a-z]+/(base/)?\d+).match?(entity_url)
```
rwd: Could we add `\A` and `\z` boundary markers to this Regexp.

app/presenters/document/field_group/section/value_presenter.rb
```diff
@@ -72,20 +75,13 @@ def linkable_for_entity?
...
-            entity_uri = URI.parse(entity.fetch('about'))
-            return nil unless entity_uri.host == 'data.europeana.eu'
-          rescue URI::Error
-            return nil
-          end
-
-          type, namespace, id = entity_uri.path.split('/')[1..-1]
-          return nil unless namespace == 'base'
+        def entity_url(format = 'html')
+          return nil unless  europeana_entity?
```
rwd: Remove one of the spaces after `unless`

app/presenters/document/field_group/section/value_presenter.rb
```diff
@@ -72,20 +75,13 @@ def linkable_for_entity?
...
+          portal_entity_path(entity.fetch('about'), slug: entity_url_slug(entity), format: format)
+        end
 
-          controller.entity_path(type: entities_human_type(type), id: id, slug: entity_url_slug(entity), format: 'html')
+        def europeana_entity?
+          @europeana_entity ||= for_entity? && europeana_entity_url?(entity.fetch('about'))
```
rwd: This memoisation is only effective when the stored value is `true`. It will need to be rewritten as something like:
```ruby
return @europeana_entity if instance_variable_defined?(:@europeana_entity)
@europeana_entity = for_entity? && europeana_entity_url?(entity.fetch('about'))
```

app/views/portal/show.rb
```diff
@@ -82,7 +82,6 @@ def content
           object: {
             annotations_later: true, # TODO: remove when styleguide assumes this
             creator: presenter.creator_title,
-            creators_info: presenter.creators_info,
```
rwd: The new field group still needs to be constructed and passed to the template, i.e. with: `creators: presenter.field_group(:creators)`

PR #1111, opened 2018-07-16
EC-2906 Move selected collection facet to top.

lbiedinger:


app/presenters/facet_presenter.rb
```diff
@@ -260,7 +258,7 @@ def apply_only_to_items(items, **_)
   end
 
   def apply_order_to_items(items, **_)
-    items
+    items.unshift(items.delete(items.detect { |item| facet_in_params?(facet_name, item) }))
```
rwd: This belongs in the collection facet presenter, not here.

app/presenters/facet_presenter.rb
```diff
@@ -239,7 +237,7 @@ def apply_only_to_items?
   end
 
   def apply_order_to_items?
-    true
+    false
```
rwd: Makes sense.

PR #1113, opened 2018-07-17
EC-2945 Order gallery filter options

lbiedinger:


app/views/galleries/index.rb
```diff
@@ -82,7 +82,7 @@ def gallery_topic_options
     end
 
     def published_gallery_topic_options
-      Topic.with_published_galleries.map do |topic|
+      Topic.with_published_galleries.order(:label).map do |topic|
```
rwd: This does not take into account that labels are translated. It needs to.

PR #1124, opened 2018-08-03
EC-2932 Record ID for image errors on galleries

lbiedinger:


app/models/gallery_image/http_response.rb
```diff
@@ -14,7 +14,7 @@ module HTTPResponse
 
     def validate_http_image
       unless http_image_content_type?
-        errors.add(:url, %(HTTP response is not an image for "#{url}"))
+        errors.add(:url, %(HTTP response is not an image for "#{url}" #{europeana_record_id}))
```
rwd: Could we add parentheses or brackets or some other punctuation to demark the record ID, to make it a bit easier to scan?

PR #1136, opened 2018-08-16
Observe pjax:end event to make tabs draggable

rwd:


app/assets/javascripts/rails_admin/custom/ui.js
```diff
@@ -11,4 +11,7 @@ $(function() {
       });
     }
   });
-});
+}
+
+$(document).ready(makeHasManyAssociationTabsDraggable);
```
lbiedinger: `$(makeHasManyAssociationTabsDraggable);` would be enough here. See https://api.jquery.com/ready/

PR #1137, opened 2018-08-16
dcterms:isPartOf promo card

rwd:


app/controllers/concerns/europeana/search_api_consumer.rb
```diff
@@ -39,6 +39,19 @@ def search_results_for_europeana_ids(europeana_ids, **options)
...
     end
 
+    # Queries the Search API for Europeana records having another as part of them
+    #
+    # @param europeana_id [String] Europeana record ID to search for records
+    #   having is as part of them
+    # @return see Europeana::API::Record.search
+    def search_results_for_dcterms_has_part(europeana_id)
+      search_options = {
+        api_url: api_url, rows: blacklight_config.default_per_page,
```
lbiedinger: this will retrieve 12 records, if we're only interested in 1 parent item perhaps this method should take per_page as an option. Then it can still default to this.

app/views/portal/parent.json.jbuilder
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+json.url document_path(@parent['id'][1..-1])
+json.relation 'Parent item'
+json.title @parent['title']
+# json.attribution @parent['dataProvider']
```
lbiedinger: can be removed if we won't be showing this anywhere.

app/controllers/portal_controller.rb
```diff
@@ -97,6 +97,15 @@ def galleries
     end
   end
 
+  def parent
+    # Search the API for the record with dcterms:hasPart data.europeana.eu/item/RECORD_ID
+    @parent = search_results_for_dcterms_has_part(doc_id)[:items]&.first
+
+    respond_to do |format|
+      format.json { render :parent, layout: false }
```
lbiedinger: should we be concerned with a case where the parent couldn't be found? For example for a record that says it's part _of some record that doesn't contain that record as a hasPart.

app/views/portal/parent.json.jbuilder
```diff
@@ -0,0 +1,11 @@
...
+
+json.url document_path(@parent['id'][1..-1])
+json.relation 'Parent item'
+json.title @parent['title']
+json.excerpt do
+  json.short Europeana::API::Record::LangMap.localise_lang_map(@parent['dcDescriptionLangAware'])
+end
+json.img do
+  json.src @parent['edmPreview']
+end
```
lbiedinger: This will need to account for:
```

is_image: true/false
is_audio: true/false
is_text: true/false
is_video: true/false
```

PR #1138, opened 2018-08-16
Entity promo card(s)

rwd:


app/models/concerns/edm/entity/depiction.rb
```diff
@@ -17,7 +17,7 @@ def thumbnail_src
       end
 
       def thumbnail_full
-        api_response[:depiction][:id]
+        api_response.key?(:depiction) ? api_response[:depiction][:id] : nil
```
lbiedinger: for this we could use:
`api_response.dig(:depiction, :id)`

https://ruby-doc.org/core-2.3.0/Hash.html#method-i-dig

app/views/entities/promo.json.jbuilder
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# presenter = EntityPresenter.new(@entity)
+
+json.url portal_entity_path(@entity.uri, format: 'html')
+json.title false
```
lbiedinger: why set this? Is there anything more applicable that can go into the title than 'false'?

app/views/entities/promo.json.jbuilder
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# presenter = EntityPresenter.new(@entity)
```
lbiedinger: Is this a leftover commented out line or supposed to document something?

PR #1141, opened 2018-08-23
Docker Compose config

rwd:


lib/commands/europeana/portal/docker/templates/minio-s3-policy.json
```diff
@@ -0,0 +1,20 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [{
+    "Action": ["s3:GetBucketLocation", "s3:ListBucket"],
+    "Effect": "Allow",
+    "Principal": {
+      "AWS": ["*"]
+    },
+    "Resource": ["arn:aws:s3:::%<= s3_bucket %>"],
```
lbiedinger: this is just coming out as a literal

lib/commands/europeana/portal/docker/templates/minio-s3-policy.json
```diff
@@ -0,0 +1,20 @@
...
+    "Sid": ""
+  }, {
+    "Action": ["s3:GetObject"],
+    "Effect": "Allow",
+    "Principal": {
+      "AWS": ["*"]
+    },
+    "Resource": ["arn:aws:s3:::%<= s3_bucket %>/*"],
```
lbiedinger: same here. needs to be properly interpolated.

lib/commands/europeana/portal/docker/docker_command.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      source_root File.expand_path('templates', __dir__)
+
+      argument :rails_env, type: :string, default: 'development', desc: 'Rails env'
+
+      class_option :europeana_api_key, type: :string, desc: 'Europeana API key'
+      class_option :redis_db_index, type: :numeric, default: 0, desc: 'Redis database index'
+
+      def copy_docker_compose
+        template 'docker-compose.services.yml', 'docker-compose.services.yml'
```
lbiedinger: why not just have the file be 'docker-compose.yml'? Then it wouldn't be necessary to add the -f flag when calling 'docker-compose xyz'

lib/commands/europeana/portal/docker/templates/docker-compose.services.yml
```diff
@@ -0,0 +1,19 @@
+version: "3"
+services:
+  minio:
+    image: minio/minio:RELEASE.2018-09-01T00-38-25Z
+    ports:
+    - "3001:9000"
```
lbiedinger: the ports here are the same as used by the europeana contribute docker services. While not necessarily a problem, since it's easy to start/stop the relevant docker instances, it might be nice to run everything here 10 or 100 ports up so that both applications can easily be run side by side.

README.md
```diff
@@ -5,25 +5,31 @@
...
 * Download the source code
 * Run `bundle install`
 
+## Quick Start with Docker
+
+* Install the gem bundle: `bundle install`
+* Generate Docker configuration files: `bundle exec bin/portal dockerize development`
+  You will be prompted to enter your Europeana API key.
+* Bring it up: `docker-compose -f docker-compose.services.yml up`
+* Setup the database: `bundle exec rake db:setup`
```
lbiedinger: either here or down below in the test section add instructions for setting up the test db.
All I needed to do was after the above commands, copy .env.development to .env.test and change the DB name. Then run: 'RAILS_ENV=test bundle exec rake db:setup'

The copying of the .env file is something the dockerize script could already do perhaps?

lib/commands/europeana/portal/docker/docker_command.rb
rwd: Because of the potential use of Docker Compose to also run the app itself, which I would envisage being the default action, i.e. using `docker-compose.yml`. But as we are not doing so yet, I will rename the file as you suggest.

lib/commands/europeana/portal/docker/templates/docker-compose.services.yml
rwd: Good idea.

lib/commands/europeana/portal/docker/docker_command.rb
```diff
@@ -0,0 +1,89 @@
...
+      def copy_docker_env
+        template '.env.docker', '.env.docker'
+      end
+
+      def copy_minio_policy
+        template 'minio-s3-policy.json', minio_policy_path
+      end
+
+      def copy_rails_env
+        template '.env.rails', ".env.#{rails_env}"
```
lbiedinger: could we append the values rather than overwriting the entire file? That way anything already specified in the .env file won't be lost, but since the last assignment will take precedence the relevant settings will still be applied.

PR #1150, opened 2018-09-12
Remove edm:landingPage workaround

rwd:
As it has been fixed at the Search & Record API level.

### DIFF
--- a/app/presenters/document/web_resource_presenter.rb
+++ b/app/presenters/document/web_resource_presenter.rb
@@ -205,8 +205,8 @@ module Document
         size_unit: 'pixels',
         runtime: field_value('ebucoreDuration'),
         runtime_unit: t('site.object.meta-label.runtime-unit-seconds'),
-        attribution_plain: attribution_snippet(:text),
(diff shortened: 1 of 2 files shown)


app/presenters/document/web_resource_presenter.rb
```diff
@@ -218,17 +218,6 @@ def media_metadata
...
 
-    def attribution_snippet(format)
-      # Disabled as textAttributionSnippet and htmlAttributionSnippet have malformed URLs
-      # field_value("#{format}"AttributionSnippet")
-
-      # Temporary workaround until textAttributionSnippet and htmlAttributionSnippet
-      # are fixed upstream
-      # TODO: Remove this workaround
-      snippet = field_value("#{format}AttributionSnippet")
-      snippet&.gsub(record_presenter.field_value('europeanaAggregation.edmLandingPage'), record_presenter.edm_landing_page)
```
lbiedinger: we can remove 'record_presenter' from the attr_reader decelerations on line 16 then

PR #1154, opened 2018-10-05
EC-3023 Exhibitions promos

lbiedinger:


app/views/portal/exhibitions.json.jbuilder
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+content = exhibition_content(@exhibition)
+json.exhibition_promo do
```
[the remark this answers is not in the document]
lbiedinger: this is the structure the frontend expects, it won't pick it up without this. I think we should leave it as is for now since the promo cards are due to be refactored anyway. Once that happens we can update this as well as remove some of the duplicate info(like 3 times the same image url being sent now).

app/controllers/portal_controller.rb
```diff
@@ -86,6 +86,22 @@ def annotations
     end
   end
 
+  # GET /record/:id/exhibition
+  def exhibition
+    creator_name = Rails.application.config.x.exhibitions.annotation_creator_name
```
rwd: Let's move lines 91-97 into a new method named `exhibition_url`

config/initializers/env.rb
```diff
@@ -34,6 +34,10 @@
   validation_mail_to: ENV['GALLERY_VALIDATION_MAIL_TO']
 )
 
+Rails.application.config.x.exhibitions = OpenStruct.new(
+  host: ENV['EXHIBITIONS_HOST'] ||'https://' + (ENV['HTTP_HOST'] || 'www.europeana.eu'),
+  annotation_creator_name: ENV['EXHIBITION_ANNOTATION_CREATOR_NAME'] || 'Europeana.eu Exhibition'
```
rwd: If this env var is going here, let's name it `EXHIBITIONS_ANNOTATION_CREATOR_NAME` (plural)

PR #1164, opened 2018-11-16
Pseudo-facet API toggle for newspapers collection

rwd:
To introduce a pseudo-facet for the Newspapers collection to switch between its custom API and the default, this:
* abstracts some of the collection-facet-specific presentation logic to be reusable across single-selection (i.e. radio button) facets, and pseudo-facets not in fact sent to the API
* breaks the `FacetPresenter` class's methods down into more modules
* adds an "api" facet to the "newspapers" collection
* skips the use of a per-collection API in the collection search if this facet has the value "default" selected (as in default API, not default radio button)
* outputs URL paths instead of full URLs for facet links, to reduce page weight

### DIFF
--- a/app/presenters/concerns/facet/item_display.rb
+++ b/app/presenters/concerns/facet/item_display.rb
@@ -19,16 +19,16 @@ module Facet
     def items_to_show_and_hide(**options)
       options.reverse_merge!(count: 5)
 
-      unhidden_items = []
-      hidden_items = items_to_display
+      show = []
+      hide = items_to_display
 
       unless facet_config.single
-        hidden_items.select { |item| facet_in_params?(facet_name, item) }.each do |selected_item|
-          unhidden_items << hidden_items.delete(selected_item)
+        hide.select { |item| facet_in_params?(facet_name, item) }.each do |selected_item|
+          show << hide.delete(selected_item)
         end
       end
-      unhidden_items.push(hidden_items.shift) while (unhidden_items.size < options[:count]) && hidden_items.present?
-      [unhidden_items, hidden_items]
+      show.push(hide.shift) while (show.size < options[:count]) && hide.present?
+      [show, hide]
     end
 
     # @return [Array<Europeana::Blacklight::Response::Facets::FacetItem>]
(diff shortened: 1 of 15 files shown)


app/presenters/concerns/facet/url.rb
```diff
@@ -0,0 +1,97 @@
...
+    def remove_facet_url(item)
+      rewrite_facet_url(:remove, item)
+    end
+
+    def replace_facet_url(item)
+      rewrite_facet_url(:replace, item)
+    end
+
+    def rewrite_facet_url(action, item)
+      path = send(:"#{action}_facet_path", item)
```
lbiedinger: all these actions are the same, no? Why not just call search_action_path here?

spec/features/newspapers_collection_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+
+RSpec.feature 'Newspapers collection' do
+  let(:collection) { collections(:newspapers) }
+  include_context :collection_with_custom_api_url
+
+  describe 'search' do
+    it 'has an API toggle' do
+      visit collection_path(collection, locale: 'en', q: '')
+
+      expect(page).to have_css('.filter-list a.is-checked .filter-text', text: 'Fulltext')
```
lbiedinger: Let's use the I18n value here and on the line below already. So in case the text changes it won't break our tests.

app/presenters/concerns/facet/item_display.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    def apply_format_value_as_to_items?
+      facet_config.format_value_as.present?
+    end
+
+    def apply_only_to_items(items)
+      items.select { |item| facet_config.only.call(item) }
+    end
+
+    def apply_order_to_items(items)
```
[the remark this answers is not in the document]
rwd: `Facet::CollectionPresenter` overrides and uses it.

app/presenters/concerns/facet/url.rb
rwd: So that sub-classes may override one or more of them.

PR #1181, opened 2018-12-03
Add more items query for new similar items

rwd:


app/views/portal/show.rb
```diff
@@ -235,7 +235,8 @@ def suggestions
     def suggestions_similar_items
       {
         tab_title: t('site.object.items-similar-to-item'),
-        url: document_similar_url(document, format: 'json', mlt_query: @mlt_query, per_page: 12)
+        url: document_similar_url(document, format: 'json', mlt_query: @mlt_query, per_page: 12),
+        more_items_query: search_path(mlt: document.id)
```
lbiedinger: shouldn't this be ```more_items_query: search_path(params.slice(:api_url).merge(mlt: document.id))``` as we were doing for similar_items previously?

rwd: I don't think it's necessary. We don't generally attempt to preserve the explicit `api_url` query parameter across links throughout the site, so no particular need to here.

PR #1203, opened 2019-01-18
Request hits from Search API and pass to Mustache

rwd:


spec/presenters/document/search_result_presenter_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  let(:document) { bl_response.documents.first }
+  let(:basic_api_response) { JSON.parse(api_responses(:search)) }
+  let(:api_response) { basic_api_response }
+
+  describe '#hit_selector' do
+    subject { described_class.new(document, controller, api_response).hit_selector }
+
+    context 'with hit in API response' do
+      let(:api_response) do
+        basic_api_response.tap do |response|
```
lbiedinger: since we'll be adding the 'hits' profile by default to all searches I'd consider including it on the basic api response as opposed to modifying it on the fly here.

rwd: But on the other hand, only the newspapers API will return hits...

PR #1214, opened 2019-01-25
Use multi_json with yajl

rwd:
And install derailed_benchmarks in profiling env


config/initializers/active_support.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+require 'active_support/number_helper/number_to_rounded_converter'
+
+module ActiveSupport
+  module NumberHelper
+    class NumberToRoundedConverter
+      # Patch to quiet warnings for deprecated `BigDecimal.new` in Ruby 2.6
+      def calculate_rounded_number(multiplier)
```
lbiedinger: Is it worth overwriting like this? If these are only warnings would it not make sense to leave them until a version of ActiveSupport is available that doesn't raise these anymore and then simply upgrading?

PR #1227, opened 2019-03-29
Hide europeana IIIIF page webResources from the media carousel 

lbiedinger:


app/presenters/document/web_resource_presenter.rb
```diff
@@ -256,13 +256,19 @@ def displayable_tests
...
+        -> { for_has_view? && europeana_iiif? ? false : nil },
         # TRUE if for a hasView and MIME type is known
         -> { for_has_view? && mime_type.present? ? true : nil },
         # FALSE otherwise
         -> { false }
       ]
     end
 
+    def europeana_iiif?
+      field_value('dctermsIsReferencedBy') =~ %r(^http(s)?://iiif\.europeana\.eu/presentation/)
```
rwd: As the scheme is always `https://` and this will not change to `http://`, let's make this:
```ruby
start_with?('https://iiif.europeana.eu/presentation/')
```

